PR #12, opened 2023-04-27
API env_package change from datasets to types

HughParry:
Merging changes made for previous env_package repo to monorepo


forgetso (commented):
LGTM

PR #14, opened 2023-04-27
Issue 44 - Optimise imports to reduce bundle size

HughParry:


HughParry:
Review should wait until the env branch is reviewed and merged with develop

goastler (commented):
First off, good job, this cleans up a lot of import/export stuff which is really nice.

Second, this is _almost_ what I was after, but perhaps I should clarify. When I was investigating webpack, I discovered that webpack can produce a smaller bundle size using tree shaking (https://webpack.js.org/guides/tree-shaking/) if the imports are in the correct format. By correct format, I remember reading that `import x from "y"` can be eliminated by webpack if `x` is not used, but `import {x} from "y"` does not have that ability, so it cannot strip the import because it cannot detect unused code for `x`, therefore the bundle size isn't reduced as much as it could be. This ticket was made to investigate that and change over the imports where possible/appropriate to reduce bundle size

goastler (approved):

HughParry:
Closing this since last commits from a month ago and likely out of date. Will optimise imports in the bundling PRs

PR #61, opened 2023-05-15
Issue/46 Automatically reselecting with a max retry limit

HughParry:


packages/api/src/api/ProviderApi.ts
```diff
@@ -30,18 +32,30 @@ export default class ProviderApi extends HttpClientBase {
         return this.axios.get(`/v1/prosopo/providers`)
     }
 
-    public getCaptchaChallenge(
+    public async getCaptchaChallenge(
```
goastler, line 33: have you changed everywhere this is used to use `await`?

HughParry: I believe it's onlyused in getCaptchaChallenge() in procaptcha/src/modules/ProsopoCaptchaApi

packages/api/src/api/ProviderApi.ts
```diff
@@ -30,18 +32,30 @@ export default class ProviderApi extends HttpClientBase {
...
+        try {
+            return this.axios.get(
+                `/v1/prosopo/provider/captcha/${provider.datasetId}/${userAccount}/${
+                    this.network.dappContract.address
+                }/${blockNumber.toString().replace(/,/g, '')}`
+            )
+        } catch (e) {
+            console.log(e)
+            const newRandomProvider = await this.getNewRandomProvider(userAccount)
+            return this.getCaptchaChallenge(userAccount, newRandomProvider, dappAccount, retryAttempt + 1)
```
goastler: you're making this function recursive here. I think it would be better to do this in a loop rather than having the retry parameter. E.g.
```
for retry in 0..10 {
  challenge = getChallenge()
  if challenge then return
}
throw new Error("max retries exceeded")
```

HughParry: Good shout, this avoids global consts :)

packages/api/src/api/ProviderApi.ts
```diff
@@ -74,4 +88,8 @@ export default class ProviderApi extends HttpClientBase {
         }
         return this.axios.post(`/v1/prosopo/provider/verify`, payload)
     }
+
+    private getNewRandomProvider(userAccount: string): Promise<ProsopoRandomProvider> {
```
goastler: is this being called in multiple places? If not, you could inline it like the other axios calls

packages/api/src/api/ProviderApi.ts
goastler: I wonder whether we should have some kind of sleep function here? E.g. sleep(1s). If the retry limit is 10 and the block time is 6s, I can see 10 retries all being done in the same block, all getting the same provider who isn't responding. If we add some kind of delay, e.g. 1 second, we would hit a block rollover and get another provider. Could do with discussing how long this delay should be / how many retries should be done.

I think we should do a fair number of retries but attempt to hit at least 3 blocks. That would give us 3 different providers, in theory. Although maybe 5 would be better.

goastler (changes requested):
Could do with discussing some of this

HughParry:
fixes prosopo/captcha-private#818

packages/api/src/api/ProviderApi.ts
forgetso: Yes agree. Block interval can be gotten from the API consts in polkadot

forgetso: `api.consts.babe?.expectedBlockTime`

forgetso: https://github.com/prosopo/captcha/blob/be7e880acbc4f70108049ad6fca6ff0dde6ab355/packages/provider/tests/batch/commitments.test.ts#L41-L60

PR #79, opened 2023-05-23
Image scraping and processing

HughParry:


goastler (dismissed):
Few more comments on the processor.py would be appreciated just so if someone else works on this they know what's going on :)

HughParry:
Migrated to other repo

PR #84, opened 2023-05-26
Issue/75 update captcha to latest version of contract with provider commit

### LINKED ISSUE #2: change package jsons to point to monorepo for all packages

forgetso:
# Completed 
- [x] Updated packages to latest version of protocol [48321f627ee7320c69287dcc94e4071f474a5983](https://github.com/prosopo/protocol/commit/48321f627ee7320c69287dcc94e4071f474a5983)
- [x] Moved eslint config to root of repo and deleted all individual package configs.
- [x] Fixed all references in tsconfig.json files
- [x] dev package now has all packages as references so `npm run b dev` will build everything
- [x] Package dependencies fixed where packages were included but not needed and vice versa
- [x] Polkadot dependencies upgraded to latest
- [x] Tests now run from from dev using command `"test": "npm -w @prosopo/dev run test",` from root of the workspace
- [x] Workflow updated to run tests from dev
- [x] Tests are now discovered in every package
- [x] Prosopo imports reverted to relative throughout packages until different build process implemented (not tsc, which does not respect paths)
- [x] TypeScript updated throughout
- [x] Typechain implemented for generating contract sources. Generated files can be imported using this command from the root of the package
```
npm -w @prosopo/dev run import:contract -- --in ../protocol/target/ink/captcha --out ./packages/contract/src/typechain
```
- [x] Types need to be copied manually into the types package in `packages/types/src/contract/typechain/captcha/types-arguments` and `packages/types/src/contract/typechain/captcha/types-returns`. These are currently duplicated in contract and types to prevent circular dependencies.
- [x] Types added to the database API URLs and API parameters to ensure that names don't go out of sync

# Todo
- [x] All tests passing
- [x] Check frontend is working
- [x] Store protocol sources in monorepo and deploy before tests or create a new substrate node with contract predeployed

# Follow Up
- [x] prosopo/captcha-private#729 
- [x] prosopo/captcha-private#728 
- [x] prosopo/captcha-private#727
- [x] prosopo/captcha-private#726
- [x] prosopo/captcha-private#725


goastler (dismissed):
lgtm

dev/src/setup/dapp.ts
```diff
@@ -9,13 +10,13 @@ export async function setupDapp(env: ProsopoEnvironment, dapp: IDappAccount): Pr
         const tasks = new Tasks(env)
 
         try {
-            await tasks.contractApi.getDappDetails(dapp.contractAccount)
+            ;(await tasks.contract.query.getDappDetails(dapp.contractAccount)).value.unwrap().unwrap()
```
HughParry, line 13: defensive semi colon as first line in try block? not sure i follow what that syntax does :)

dev/src/setup/setup.ts
```diff
@@ -108,7 +109,7 @@ export async function setup() {
         const env = new Environment(pair, defaultConfig())
         await env.isReady()
 
-        const dappStakeDefault = (await env.contractInterface.getDappStakeDefault()).muln(2)
+        const dappStakeDefault = (await env.contractInterface['dappStakeThreshold']()).muln(2)
```
HughParry, line 112: is this a Typescript thing or have the types/interfaces changed?

packages/provider/tests/batch/commitments.test.ts
```diff
@@ -125,22 +131,33 @@ describe('BATCH TESTS', function () {
                 const commitmentIds: string[] = []
 
                 // Store 10 commitments in the local db
+                const completedAt = (await env.api.rpc.chain.getBlock()).block.header.number.toNumber()
+                const requestedAt = completedAt - 1
```
HughParry, line 135: not sure if i understand why requestedAt = completedAt - 1. what if it takes longer than 1 block to complete?

HughParry (approved):

dev/src/setup/dapp.ts
forgetso: It was auto completed by eslint presumably for [this](https://stackoverflow.com/a/10044845) reason and TS won't compile without it.

dev/src/setup/setup.ts
forgetso: The function `getDappStakeDefault` no longer exists as we're now using typechain-polkadot to generate the contract interface for functions. `dapp_stake_threshold` is a storage field in the contract and I've added a getter by the same name to the contract:

https://github.com/prosopo/captcha/blob/007cde247e94e2e79f95e5c832abc856ae9c7213/packages/contract/src/contract/interface.ts#L108-L120

I can't use `env.contractInterface.dappStakeThreshold` because TS doesn't know that this function exists as its created after compile time. In future, we'll add some fancy trick to make TS aware of storage functions.

packages/provider/tests/batch/commitments.test.ts
forgetso: This just a test to see if we can batch 50 commitments on-chain. The content of the commitments doesn't matter so I'm simply creating dummy ones here.

goastler (approved):

PR #85, opened 2023-05-30
Retry random provider within observable

HughParry:


packages/procaptcha/src/modules/Manager.ts
```diff
@@ -197,46 +198,37 @@ export const Manager = (
...
+                        updateState({
+                            index: 0,
+                            solutions: challenge.captchas.map(() => []),
+                            challenge,
+                            showModal: true,
+                            timeout,
+                            blockNumber,
+                        })
+                    },
+                    (error) => console.error('Failed to fetch challenge and block number', error)
```
goastler: this should be some kind of callback rather than silently logging to console

packages/procaptcha/src/modules/Manager.ts
```diff
@@ -197,46 +198,37 @@ export const Manager = (
...
+                )
         } catch (err) {
```
goastler, line 272: the stuff inside the defer call won't be caught by this catch now. Would it be better to put the whole of this fn inside the defer fn perhaps?

packages/procaptcha/src/modules/Manager.ts
```diff
@@ -247,6 +239,59 @@ export const Manager = (
...
+                            unsubscribe()
+                            subscriber.next()
+                        })
+                    })
+                })
+            )
+        }
+    }
+
+    const fetchCaptchaChallenge = async (contract, account, config) => {
```
goastler: good work breaking these up into fns

packages/procaptcha/src/modules/Manager.ts
```diff
@@ -247,6 +239,59 @@ export const Manager = (
...
+                    if (retryAttempt > maxRetryAttempts) {
+                        return throwError(
+                            () => new Error(`No responsive provider found after ${maxRetryAttempts} attempts`)
+                        )
+                    }
+                    // Wait for the next block
+                    return new Observable((subscriber) => {
+                        const unsubscribe = api.rpc.chain.subscribeNewHeads((header) => {
+                            console.log(`Chain is at block: #${header.number}`)
+                            unsubscribe()
```
goastler: that's some gnarly subscriptions there haha

packages/procaptcha/src/modules/Manager.ts
```diff
@@ -482,3 +527,19 @@ export const Manager = (
         nextRound,
     }
 }
+function timoutSetup(
```
goastler: who's tim and why is he out? :joy:

goastler (changes requested):
need to discuss some of the changes :)

packages/procaptcha/src/modules/Manager.ts
goastler: nvm, moot point because the isHuman checks would be repeated and we don't want that to happen

HughParry:
Closing since now ancient code from when I joined.

Ticket in to resolve this here https://github.com/prosopo/captcha-private/issues/523

PR #88, opened 2023-05-30
Fix client example

### LINKED ISSUE #87: Fix client-example
Nothing is rendering in client-example due to TS not transpiling. The issue is actually because the way in which the component is exported has changed in procaptcha-react

forgetso:


dev/src/setup/provider.ts
```diff
@@ -72,6 +72,8 @@ export async function setupProvider(env: ProsopoEnvironment, provider: IProvider
     } catch (e) {
         logger.warn(e)
     }
+    const registeredProvider = await env.contractInterface.getProviderDetails(provider.address)
+    console.log(registeredProvider)
```
goastler: left over debug?

goastler (dismissed):

dev/src/setup/provider.ts
forgetso: yep, removed

HughParry (approved):

PR #94, opened 2023-05-31
Ts-node instead of node

### LINKED ISSUE #93: change clis and pkg jsons to use ts-node instead of node

goastler:
fixes prosopo/captcha-private#795


forgetso:
I tried running ts-node with client-example-server and it was slower to start than with the tsc build + node run command. Plus, ts-node [does not watch file changes](https://typestrong.org/ts-node/docs/recipes/watching-and-restarting/) and contains no build output. What is the benefit of using ts-node if it's not watching file changes? It has to build the files from source each time it's run, which means it would be slower than running the normal `.js` command if there were file changes.

goastler:
defunct

PR #135, opened 2023-06-08
Rust keep build artifacts

goastler:


goastler:
fixes #135

.gitignore
```diff
@@ -119,7 +119,10 @@ secrets.json
 .env*
 
 # Ingnore build target
-contracts/target/
+protocol/target/ink/*/release
+protocol/target/ink/*/wasm32-unknown-unknown
```
forgetso, line 123: Is it always wasm32-unknown-unknown?

forgetso (changes requested):

.gitignore
goastler: yes it is, that's the target we're aiming at. Ink always aims at wasm32-unknown-unknown because that's the format expected for contract code on the blockchain. I doubt they'll change it anytime soon. Can potentially see it becoming wasm64 one day, but certainly not soon

forgetso (approved):

Issue #121, opened 2023-06-07
add an env variable for specifying the author of the contract when building the contract


goastler:

goastler:
chris says cargo contract should forward the env variables

goastler:
register this as an issue with cargo contract, but for the time being make a js find and replace util for the contract to find and replace the author field before a build, and call this in the cli before build

forgetso:
Proper fix should be requested from cargo contract.

PR #140, opened 2023-06-09
Protocol env

goastler:
Allow environment variables to be populated in our ink contracts.

Environment variables should start with `INK_`, e.g. `INK_AUTHOR`.

This is a somewhat rough-around-the-edges way of doing env variables. It relies on a regex find and replace. There's many things that could go wrong in this process if someone tried to fiddle with it, but it'll do for now until cargo-contract can handle env variables.


goastler:
fixes #121

goastler:
Just need to change the js side to match these changes, minor though

forgetso (approved):

PR #142, opened 2023-06-13
97 enable cypress testing

### LINKED ISSUE #97: enable cypress testing

### LINKED ISSUE #123: fast-xml-parser audit issue

forgetso:
Enable some basic cypress testing as part of provider workflow.


packages/procaptcha/src/modules/Manager.ts
```diff
@@ -122,11 +118,9 @@ export const Manager = (
     const start = async () => {
         try {
             if (state.loading) {
-                console.log('Procaptcha already loading')
```
goastler, line 125: would it be better to move these to console.debug()s?

also, I believe webpack can strip out the console logs so we don't have to ourselves

goastler (changes requested):

packages/procaptcha/src/modules/Manager.ts
forgetso: Totally. Yeah, we can strip console logs using webpack and we should do that.

goastler: discussed on slack

goastler (approved):

PR #166, opened 2023-06-20
Contract sources update

goastler:
Update the js code to the latest version of the contract


packages/provider/tests/tasks/tasks.test.ts
```diff
@@ -321,27 +321,27 @@ describe('CONTRACT TASKS', async function (): Promise<void> {
...
+        const provider = (await tasks.contract.query.getProvider(accountAddress(providerAccount))).value
             .unwrap()
             .unwrap()
 
         const completedAt = (await env.api.rpc.chain.getBlock()).block.header.number.toNumber()
         const requestedAt = completedAt - 1
         const providerTasks = await getSignedTasks(env, providerAccount)
         await providerTasks.contract.tx.providerCommit({
-            dapp: accountContract(dappContractAccount),
+            dappContract: accountContract(dappContractAccount),
```
forgetso, line 332: Is there a reason why these have all changed back? I think you need to cd into provider and run `npm run build:test`

goastler:
fixes #121

packages/types/src/contract/typechain/captcha/types-returns/captcha.ts
```diff
@@ -34,34 +53,39 @@ export enum Error {
...
+export type UserHistorySummary = {
+	correct: number,
+	incorrect: number,
+	score: number
 }
 
 export type Commit = {
 	id: Hash,
-	user: AccountId,
+	userAccount: AccountId,
```
forgetso, line 84: Looks like we're both trying to accommodate for the other part. I had changed everything in TS to dapp/provider/user to match the contract.

forgetso (changes requested):

packages/types/src/contract/typechain/captcha/types-returns/captcha.ts
goastler: let's chat on slack :)

forgetso (dismissed):

forgetso (approved):

PR #178, opened 2023-06-21
Inline ink env variables

goastler:
allow for const or let as declaration for env variables in the contract inline rather than as consts only


HughParry (dismissed):

goastler:
fixes #179

HughParry (approved):

PR #183, opened 2023-06-22
171 mimic recaptcha in procaptcha bundle

forgetso:


packages/procaptcha-react/src/components/ExtensionAccountSelect.tsx
```diff
@@ -35,9 +35,11 @@ export const ExtensionAccountSelect = ({
 
     useEffect(() => {
         const prom = web3Enable(dappName).then(() => {
+            console.log('subscribe')
```
goastler: probably don't need this and if we do it should be more explicit as to what it's subscribing to

packages/procaptcha-react/src/components/ExtensionAccountSelect.tsx
```diff
@@ -35,9 +35,11 @@ export const ExtensionAccountSelect = ({
...
             return web3AccountsSubscribe(setAccounts)
         })
         return () => {
+            console.log('unsubscribe')
```
goastler: probably don't need this and if we do it should be more explicit as to what it's subscribing to

packages/procaptcha/src/api/ExtensionWeb2.ts
```diff
@@ -79,7 +82,9 @@ export default class ExtWeb2 extends Extension {
         const browserEntropy = await this.getFingerprint()
         const canvasEntropy = picassoCanvas(params.numberOfRounds, params.seed, params)
         const entropy = hexHash([canvasEntropy, browserEntropy].join(''), 128).slice(2)
-
+        console.log('canvas entropy', canvasEntropy)
```
goastler: do we need these console logs?

packages/procaptcha/src/api/sign.ts
```diff
@@ -6,10 +6,10 @@ export function sign(message: string, account): { messageHash: string; signature
     const keyring = new Keyring({ type: 'ecdsa' })
     const pair = keyring.addFromUri('//Alice')
     const sig = pair.sign(message)
-
+    console.log('sig', sig)
```
goastler: do we need these console logs?

packages/procaptcha/src/modules/Manager.ts
```diff
@@ -100,11 +118,11 @@ export const Manager = (
      * @returns the config for procaptcha
      */
     const getConfig = () => {
-        const config: ProcaptchaConfig = {
-            web2: false,
+        const config: ProcaptchaClientConfig = {
             userAccountAddress: '',
             ...configOptional,
         }
+        console.log(config)
```
goastler: do we need these console logs? / should be more explicit

packages/types/src/config/config.ts
```diff
@@ -0,0 +1,181 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { LogLevel } from '@prosopo/common'
+import { z } from 'zod'
+
+export enum DatabaseTypes {
+    mongo = 'mongo',
```
goastler: I think enums like this can be presented differently to avoid the `name = str` pattern, which can lead to accidental inconsistencies, plus we have to type everything twice.

I'm thinking something along the lines of `type DatabaseTypes = 'mongo' | 'mongoMemory';`, thoughts? afaik this is recommended ahead of enum usage in typescript. Given we're not mapping an enum name to a _different_ string key, the `type` union makes more sense imo

packages/types/src/config/config.ts
```diff
@@ -0,0 +1,181 @@
...
+})
+
+export const ProsopoCaptchaSolutionConfigSchema = z.object({
+    requiredNumberOfSolutions: z.number().positive().min(2),
+    solutionWinningPercentage: z.number().positive().max(100),
+    captchaFilePath: z.string(),
+    captchaBlockRecency: z.number().positive().min(2),
+})
+
+export const ProsopoServerConfigSchema = ProsopoBasicConfigSchema.merge(
```
goastler: can we make `ProsopoServerConfigSchema` and `ProsopoClientConfigSchema` inherit from a shared config at all? It can be private. There's a lot of shared fields between the two of them so it makes sense to have a single declaration, especially considering they need to be the same as each other (e.g. one shouldn't have `userAccountAddress` while the other has `userAccount` both using those fields for the same thing, we should force them to be the same name for consistency, hence shared fields)

This is why I used the omit function to remove a couple of fields from the other type, however I prefer these types with a shared set of fields

goastler (changes requested):

packages/types/src/config/config.ts
forgetso: My IDE auto completes enums like this for me.

I've read [this](https://stackoverflow.com/a/76485953) and it seems like `z.enum` is the way to go. I didn't realise there were issues with native enums.

The main thing I want to avoid is having to write strings everywhere like `'mongo'`, as these aren't picked up when refactoring.

packages/types/src/config/config.ts
forgetso: Done

goastler (dismissed):

goastler (approved):

PR #190, opened 2023-06-26
Style the paragraph instead

forgetso:


packages/procaptcha-react/src/components/Procaptcha.tsx
```diff
@@ -137,7 +137,7 @@ export const Procaptcha = (props: ProcaptchaProps) => {
                 )}
             </Backdrop>
 
-            <Box p={1} sx={{ maxWidth: '600px', minWidth: '200px' }} data-cy={'button-human'}>
```
goastler, line 140: I think we need to keep maxWidth. Have you seen what happens without it?

goastler (changes requested):

goastler (approved):

PR #198, opened 2023-07-03
Implement checks to see if tasks are running before spawning new ones

forgetso:


packages/provider/src/tasks/calculateSolutions.ts
```diff
@@ -0,0 +1,84 @@
...
+                    .unwrap()
+                    .unwrap()
+
+                // Get any unsolved CAPTCHA challenges from the database for this datasetId
+                const unsolvedCaptchas = await this.db.getAllCaptchasByDatasetId(
+                    provider.datasetId.toString(),
+                    CaptchaStates.Unsolved
+                )
+
+                // edge case when a captcha dataset contains no unsolved CAPTCHA challenges
```
HughParry, line 32: Redundant check since additionally checked on line 45

packages/provider/src/tasks/calculateSolutions.ts
```diff
@@ -0,0 +1,84 @@
...
+export class CalculateSolutionsTask extends Tasks {
+    constructor(env: ProviderEnvironment) {
+        super(env)
+        this.logger = logger(env.config.logLevel, 'CalculateSolutionsTask')
+    }
+
+    /**
+     * Apply new captcha solutions to captcha dataset and recalculate merkle tree
+     */
+    async run(): Promise<number> {
```
HughParry, line 17: feels like a pretty big function, but also it's pretty clear what it does. not sure if worth breaking down into smaller methods or not

packages/provider/src/util.ts
```diff
@@ -104,3 +105,18 @@ export function updateSolutions(solutions: pl.DataFrame, captchas: Captcha[], lo
         return captcha
     })
 }
+
```
HughParry, line 108: Not sure I understand this logic. If the task was running, wouldn't task.status=running, not completed?

packages/provider/src/tasks/calculateSolutions.ts
forgetso: It's going to be separated into a completely different package for downstream processing so probably best to leave refactoring till then.

packages/provider/src/util.ts
forgetso: As discussed on slack, this logic was wrong. A unit test has been added for the function

packages/provider/src/tasks/calculateSolutions.ts
forgetso: Nice spot, removed.

forgetso: I removed the check but it's required by TypeScript so I added it back in...

HughParry (approved):

PR #200, opened 2023-07-03
added storage key to captcha contract

goastler:


goastler:
This is what Hernando advised doing to make sure the contract storage is always pointing to the same location irrelevant of the contract. This works fine, I've tested it, but there are additional hurdles to jump when migrating contract storage as I've put in slack

HughParry (approved):

PR #202, opened 2023-07-04
Scripts for creating demo datasets

HughParry:
in case we want these


HughParry:
This is all now done in ts

PR #279, opened 2023-07-18
265 pass through callback handlers to procaptcha

forgetso:


packages/procaptcha-react/src/components/Procaptcha.tsx
```diff
@@ -158,6 +158,23 @@ const useProcaptcha = (): [ProcaptchaState, ProcaptchaStateUpdateFn] => {
     ]
 }
 
+function getCSSAttribute(defaultValue: string, cssAttr?: string): string {
+    if (cssAttr) {
+        // check if config.maxWidth is a valid css value for maxWidth
+        for (const postfix of ['px', '%', 'em', 'rem', 'ch', 'vw', 'vh', 'vmin', 'vmax']) {
```
HughParry: There are loads of other legal css values for length - happy to just say we only support these ones though

https://developer.mozilla.org/en-US/docs/Web/CSS/length

HughParry (dismissed):

goastler (approved):
lgtm :rocket:

packages/procaptcha-react/src/components/Procaptcha.tsx
forgetso: have decided to pass through mui theme instead.

PR #346, opened 2023-08-07
Image server

goastler:


goastler:
fixes prosopo/captcha-private#609

packages/file-server/env.example
```diff
@@ -0,0 +1,5 @@
+REMOTES="http://localhost:3001" # list of remote servers to proxy requests to (can be JSON array)
```
forgetso, line 1: I hi

packages/file-server/env.example
```diff
@@ -0,0 +1,5 @@
...
+PORT=3000 # port to run the server on
```
forgetso: I think the default port should be in line with the rest of our ports. We've used `9228` to `9231` so this one should perhaps be `9232`

forgetso (approved):

forgetso (changes requested):
Looks good. Just one very small env file query.

packages/file-server/env.example
goastler: good shout, I agree :)

forgetso (approved):

PR #369, opened 2023-08-10
remove name from cargo tomls, no longer required

goastler:


goastler:
fixes prosopo/captcha-private#560

forgetso (approved):

PR #372, opened 2023-08-10
Npm scripts

goastler:
Cleaned up the npm scripts for the workspace. Let me know your thoughts

Main points:
- removed git commands as we no longer use submodules
- removed old / incompatible commands
- removed build+deps ability as we've got the tsconfig refs straightened out now so builds know their dependencies during tsc
- removed the shorthand for build / rebuild / clean, doing this for all variants of each of those commands makes npm scripts very lengthly. It should really be a shell alias at this point, that way it applies to other packages/projects too


forgetso (changes requested):
`npm run b dev` is my favourite command

forgetso (approved):

PR #384, opened 2023-08-16
Mongodo srv port fix

HughParry:


docker/docker-compose.provider.yml
```diff
@@ -31,24 +31,21 @@ services:
...
-      - "80:80"
-      - "443:443"
-      - "443:443/udp"
-    volumes:
-      - ./docker/provider.Caddyfile:/etc/caddy/Caddyfile
-      - caddy_data:/data
-      - caddy_config:/config
-    networks:
-      - external
+      - "9615:9615"
```
forgetso: Please separate this into a different docker compose file as the old one is valid for VPS setup.

forgetso (changes requested):
Separate out docker compose but otherwise all good!

docker/docker-compose.provider.yml
HughParry: done

forgetso (approved):

PR #420, opened 2023-08-22
Flux image deploy scripts

HughParry:


packages/database/src/databases/mongo.ts
```diff
@@ -155,11 +159,14 @@ export class ProsopoDatabase extends AsyncFactory implements Database {
...
             }
 
-            await this.tables?.dataset.updateOne(
-                { datasetId: parsedDataset.datasetId },
-                { $set: datasetDoc },
-                { upsert: true }
-            )
+            await this.tables?.dataset
+                .updateOne({ datasetId: parsedDataset.datasetId }, { $set: datasetDoc }, { upsert: true })
+                .catch((err) => {
```
goastler: you're printing an error but continuing?

goastler (changes requested):

packages/database/src/databases/mongo.ts
HughParry: This looks suspiciously like stuff left in while testing

goastler (approved):

PR #425, opened 2023-08-29
combine user signature parts now that ink supports 64 length arrays

goastler:


forgetso (changes requested):
Looks good but this needs changed throughout the TypeScript stuff as well.

goastler:
will sort

goastler:
fixes #463

goastler:
fixes #461

goastler:
fixes #462

goastler:
fixes #459

goastler:
fixes #458

goastler:
this has grown in size and now fixes crucial bugs with typechain and ink 4.3.0

dev/config/src/projectInfo.ts
```diff
@@ -0,0 +1,111 @@
...
+        procaptchaReact,
+        procaptchaBundle,
+        provider,
+        server,
+        typesDatabase,
+        typesEnv,
+        types,
+        typesTypechain,
+    }
+}
```
ghost, line 111: Summary:
This code change introduces two exported functions: `getContractNames` and `getPaths`. The `getContractNames` function returns an array of contract names. The `getPaths` function returns an object of paths, which are constructed based on the root path and the specific directory names.

Feedback:
1. The `getContractNames` function (lines 1-3) is straightforward and doesn't have any issues. It's a simple function that returns an array of hardcoded contract names.

2. The `getPaths` function (lines 27-111) is more complex. It constructs an object of paths based on a root path and specific directory names. The root path is determined by the URL of the current module.

Suggestions:
1. The `getPaths` function is quite long and repetitive. Consider creating a helper function to construct the paths. This function could take the root and the specific directory as arguments and return the constructed path.

2. The `Paths` type (lines 5-25) has a lot of properties. If these properties are not likely to change often, it's fine to keep them as is. However, if they are likely to change or if new properties are likely to be added, consider using a more dynamic approach to define this type.

3. The root path in `getPaths` function (line 28) is hardcoded to go three levels up from the current module. This might cause issues if the directory structure changes. Consider passing the root path as an argument to the `getPaths` function or determining it in a more dynamic way.

4. The `getPaths` function constructs paths for a lot of directories. If not all of these paths are used often, consider splitting this function into multiple smaller functions, each returning paths for a specific part of the project (e.g., one function for `demos`, another for `packages`, etc.). This would make the code more modular and easier to maintain.

goastler:
blocked by #421

HughParry (dismissed):

forgetso (approved):

PR #426, opened 2023-08-29
add cli bash script wrapper

goastler:
This allows us to do `./cli blah blah blah` instead of `npm run cli -- blah blah blah`


forgetso (dismissed):
Should these files have .sh extensions so that people know what they are or is no file type a convention for some reason?

goastler:
> Should these files have .sh extensions so that people know what they are or is no file type a convention for some reason?

I dropped the type just to make it shorter, otherwise you have to do `./cli.sh` instead of `./cli`, that's the only reason. Happy to add extensions back in?

goastler:
alternatively I can add `#! /bin/bash` at the top to specify the shell type

forgetso:
Happy to have no extensions if you put the header in the files.

HughParry (dismissed):

goastler:
> Happy to have no extensions if you put the header in the files.

done

goastler:
closing as can use bash / zsh aliases, leave up to the user/dev

PR #427, opened 2023-08-29
make image_provider workflow only trigger on main branch pushes

goastler:
Previously, the image_provider job would run for all PRs. This means that whenever the version number is bumped in a PR, the image built in that PR is pushed to docker. If subsequent changes are made to that PR, the docker image is not replaced as the version number has not changed.

The image should only be pushed to docker __after__ the PR has been merged

### DIFF
--- a/.github/workflows/image_provider.yml
+++ b/.github/workflows/image_provider.yml
@@ -6,12 +6,7 @@ on:
       - 'main'
     paths:
       - 'packages/**'
-  pull_request:
-    branches: [ main ]
-    paths:
-      - 'packages/**'
-      - '.github/workflows/image_provider.yml'
-
+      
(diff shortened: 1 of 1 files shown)


forgetso (changes requested):
This will mean that the image is not built until we're merging to main so we can't catch any problems until then. Ideally, we'd catch the build issues before we merge to main, as main should be our point of release.

goastler:
Okay in that case I think we need two separate workflows:
- build image (run on a PR)
- publish image (run on push to main)

The former detect issues with building whereas the latter publishes a working image

goastler:
closing as superseded by end-to-end workflow (#569 ) and separate docker publish workflow (#579 )

PR #471, opened 2023-09-06
470 bundle nodejs polars for provider

forgetso:


dev/config/src/vite/vite-plugin-close.ts
HughParry, line 1: is the logger not behaving itself here - loglevel issues or something?

forgetso: You get more info when you output with `console.error` for some reason.

packages/cli/src/commands/providerDeregister.ts
```diff
@@ -1,13 +1,15 @@
...
 import { Tasks } from '@prosopo/provider'
 import { validateAddress } from './validators.js'
-export default (env: ProviderEnvironment, tasks: Tasks, cmdArgs?: { logger?: Logger }) => {
-    const logger = cmdArgs?.logger || env.logger
+export default (pair: KeyringPair, config: ProsopoConfig, cmdArgs?: { logger?: Logger }) => {
+    const logger = cmdArgs?.logger || getLogger(LogLevelSchema.Values.Info, 'cli.provider_deregister')
 
     return {
         command: 'provider_deregister',
-        description: 'Deregister a Provider',
```
goastler, line 10: why have all descriptions been changed to describe?

goastler (changes requested):

packages/cli/src/commands/providerDeregister.ts
forgetso: because type checking was not working before! `describe` is the correct fields. The previous fields were `description` or `descrption` spelled incorrectly!

goastler: oh right, I thought the field was description! fair enough :)

goastler (approved):

PR #499, opened 2023-09-12
NPM publish release process

HughParry:
Pushes to main should trigger a new NPM release if there is a version bump which doesn't end in -beta


dev/scripts/env.test
```diff
@@ -9,7 +9,7 @@ DATABASE_USERNAME=root
 DATABASE_NAME=prosopo_test
 DATABASE_HOST=127.0.0.1
 DATABASE_PORT=27017
-SUBSTRATE_NODE_URL=ws://localhost:9944
+SUBSTRATE_NODE_URL=ws://host.docker.internal:9944/
```
forgetso: Will this work outside of WSL land? That's the question.

forgetso (changes requested):
One small thing to consider.

forgetso (approved):
Looks good!

PR #528, opened 2023-09-14
Typechain fix

goastler:
- bumped version of our ink fork to v4.3.0 and added sr25519 changes, as these haven't been released yet
- bumped typechain to v1.1.1 and types v1.1.2 to work with ink v4.3.0 and get around duplicate Mapping and Lazy types in output
- fixed cli to run typechain
- fixed esm imports for typechain output (e.g. add `.js` or `assert {type:"json"}` where appropriate)
- add project info fns (e.g. to list contract, list paths to pkgs, etc)


goastler:
blocked by #421

goastler:
blocked by #425

dev/scripts/src/contract/import.ts
```diff
@@ -11,12 +11,77 @@
...
+                        result = `import ${name} from '${src}.js'`
+                    }
+                    console.log(`Replacing \n\t${match}\nwith\n\t${result}\nin ${filePath}`)
+                    return `${result}`
+                })
+                // eslint-disable-next-line no-useless-escape
+                replaced = replaced.replace(/(^(?!\s*\/\/\s*eslint).*)\n+(\s*)\/\/\s*@ts-ignore/gm, (match, p1, p2) => {
+                    const result = `${p1}\n${p2}// eslint-disable-next-line @typescript-eslint/ban-ts-comment\n${p2}// @ts-ignore`
+                    console.log(`Replacing \n\t${match}\nwith\n\t${result}\nin ${filePath}`)
+                    return result
```
HughParry, line 64: Can we refactor this file when next working in this area? Would benefit from breaking up into helper functions, strings to named constants

Also could be using the logger from common if that's not gonna cause deps issues

goastler: we can, but all of this should be temporary because ideally typechain will fix everything here on there end, then we don't have to do any post-processing

forgetso (dismissed):

HughParry (approved):

PR #534, opened 2023-09-15
Dependent issues bot

goastler:


.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,59 @@
...
+    issues:
+        types:
+            - opened
+            - edited
+            - closed
+            - reopened
+    pull_request_target:
+        types:
+            - opened
+            - edited
```
HughParry, line 13: Probably doesn't need to run for edited unless there's some issues with tickets

HughParry: fine for this but if we're ever worrying about github actions minutes then we can remove stuff like this

HughParry (approved):

.github/workflows/dependent-issues.yml
goastler: this is just in case people add further "depends on #xyz" to a PR. It's counted as an edit :)

PR #535, opened 2023-09-17
Move typechain

goastler:
blocked by #528 

This relocates our contracts' typechain output into their own js packages. This is better than putting them in the types package because it's not just types output by typechain. We need 1 pkg per contract to avoid duplicate exports, e.g. each typechain run makes a `Contract` type and we'd have to rename all of these to `XyzContract` to avoid duplication.

Questions:
- where should the contract packages live? They're currently under /contracts
- what should the contract packages be called? it's currently @prosopo/xyz-contract
- what should the dir structure be? it's currently /contracts/xyz with @prosopo/xyz-contract, but this is odd because the folder is not the same as the pkg name


dev/scripts/src/setup/setup.ts
```diff
@@ -46,7 +46,7 @@ function getDefaultProvider(): IProviderAccount {
     return {
         url: process.env.API_PORT ? `http://localhost:${process.env.API_PORT}` : 'http://localhost:3000',
```
forgetso, line 47: This is nothing to do with this PR but can you change the default to 9229?

forgetso (changes requested):

dev/scripts/src/setup/setup.ts
forgetso: I'll do it in my readme PR

forgetso (approved):

PR #569, opened 2023-09-19
End-to-end workflow

goastler:


goastler:
blocked by #535 

need contract packages to be available, then can load in the contract abi + wasm from the packages directly for the correct version in e2e workflow

goastler:
had an epiphany that we can keep the tests separate for protocol and still check typechain is up to date!

forgetso (dismissed):

goastler:
blocked by #647

dev/scripts/src/contract/import.ts
```diff
@@ -81,23 +86,23 @@ async function importContract(pathToAbis: string, pathToOutput: string) {
...
+                    if (verbose) console.log(`Replacing \n\t${match}\nwith\n\t${result}\nin ${filePath}`)
                     return `${result}`
                 })
                 // eslint-disable-next-line no-useless-escape
                 replaced = replaced.replace(/\n(.*)\n(\s*)\/\/\s*@ts-ignore/g, (match, p1, p2) => {
                     // don't replace if already ignored by eslint
                     if (p1.includes('eslint-disable-next-line')) return match
                     const result = `\n${p1}\n${p2}// eslint-disable-next-line @typescript-eslint/ban-ts-comment\n${p2}// @ts-ignore`
-                    console.log(`Replacing \n\t${match}\nwith\n\t${result}\nin ${filePath}`)
+                    if (verbose) console.log(`Replacing \n\t${match}\nwith\n\t${result}\nin ${filePath}`)
```
forgetso, line 97: How about 
```
if ( !verbose ) { console.log = (x: string)=> {} }
``` 
to avoid all of these `if` statements, or are there other places where `console.log` is used when `verbose` is false?

forgetso (changes requested):

dev/scripts/src/contract/import.ts
forgetso: Actually, this might be really confusing for someone else. Ignore.

forgetso (approved):

dev/scripts/src/contract/import.ts
goastler: I added a bool to control the printing of this because it gets *very* verbose. It's more of a debugging / verbose logging feature to figure out what's not working if/when it comes to that case

PR #585, opened 2023-09-20
Add commonJS build for packages

forgetso:


dev/config/src/dependencies.ts
```diff
@@ -2,6 +2,8 @@ import { Glob } from 'glob'
 import { ProsopoEnvError, getLogger } from '@prosopo/common'
 import { at } from '@prosopo/util'
 import child_process from 'child_process'
+import fs from 'fs'
+import path from 'path'
 import util from 'util'
 const logger = getLogger(`Info`, `config.dependencies.js`)
 const exec = util.promisify(child_process.exec)
```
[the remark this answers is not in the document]
HughParry: Is the bot right to be complaining here?

I'm starting to think we should be adding the no-unused-imports rule to eslint, then disabling it where it's required for some reason

dev/config/src/dependencies.ts
```diff
@@ -23,6 +25,88 @@ async function getPackageDir(packageName: string): Promise<string> {
...
+                }
+                // remove tsconfig.*.json from the path and get the path to the new directory via the reference path
+                let refTSConfigPath = path.resolve(tsConfigPath.replace(tsConfigRegex, ''), reference.path)
+                if (!refTSConfigPath.endsWith('.json')) {
+                    refTSConfigPath = path.resolve(refTSConfigPath, 'tsconfig.json')
+                }
+                const newTsConfigs = getTsConfigs(refTSConfigPath, ignorePatterns, tsConfigs)
+                if (newTsConfigs.length > 0) {
+                    tsConfigs = [...new Set(tsConfigs.concat(newTsConfigs))]
+                }
```
HughParry: Readability here could be a bit better - maybe worth 10 min refactor if it's not temp stuff?

HughParry (dismissed):
2 quick comments :)

dev/config/src/dependencies.ts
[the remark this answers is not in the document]
forgetso: No, the imports are used. It's an error in the bot. We already have the unused imports rule in eslint and it automatically removes them whilst your coding (at least in IntelliJ).

https://github.com/prosopo/captcha/blob/967e7605e79511ab5ea814b5e0cd831234c3c09e/dev/config/src/dependencies.ts#L39

https://github.com/prosopo/captcha/blob/967e7605e79511ab5ea814b5e0cd831234c3c09e/dev/config/src/dependencies.ts#L55-L57

dev/config/src/dependencies.ts
```diff
@@ -1,11 +1,15 @@
...
+import path from 'path'
 import util from 'util'
+
 const logger = getLogger(`Info`, `config.dependencies.js`)
 const exec = util.promisify(child_process.exec)
-
+const tsConfigRegex = /\/[a-z.]*\.json$/
 const peerDepsRegex = /UNMET\sOPTIONAL\sDEPENDENCY\s+(@*[\w-/._]+)@/
 const depsRegex = /\s+(@*[\w-/._]+)@/
 async function getPackageDir(packageName: string): Promise<string> {
```
ghost: Summary:
The code is a set of utility functions for managing package dependencies in a Node.js project. It includes functions to get package directories, resolve tsconfig paths, get tsconfig paths, get workspace externals, get dependencies, filter dependencies, and get files in directories. The code is well-structured and includes helpful comments explaining the purpose of each function.

Feedback:
1. Error Handling: The code uses error handling in some places but not all. For example, in the `getPackageDir` function, if the `exec` function throws an error, it is caught and a `ProsopoEnvError` is thrown. However, in the `getDependencies` function, if `exec` throws an error, it is not caught. This could lead to unhandled promise rejections. 

2. Regex Patterns: The code uses regular expressions to match certain patterns in strings. It would be helpful to add comments explaining what these patterns are meant to match, especially for more complex patterns like `tsConfigRegex`, `peerDepsRegex`, and `depsRegex`.

3. Code Duplication: There is some code duplication in the `getDependencies` function where the same operation is performed on `deps` and `peerDeps`. This could be refactored into a separate function to reduce duplication.

4. Use of Promises: The code uses both callbacks and promises. It would be more consistent to use only one or the other. For example, in the `getExternalsFromReferences` function, a new Promise is created to handle the asynchronous `fs.stat` and `fs.readFile` operations. This could be refactored to use `util.promisify` like the `exec` function does.

Suggestions:
1. Add error handling to all asynchronous operations to prevent unhandled promise rejections.
2. Add comments explaining the purpose of regular expressions.
3. Refactor duplicated code into separate functions.
4. Refactor callback-based asynchronous operations to use promises for consistency.

dev/config/src/dependencies.ts
forgetso: I've used filter and map but kept the recursive reference within the function that's being called recursively.

packages/provider/package.json
```diff
@@ -5,12 +5,19 @@
...
+    },
     "scripts": {
         "clean": "tsc --build --clean",
-        "build": "tsc --build --verbose",
+        "build": "tsc --build --verbose tsconfig.json",
+        "build:cjs": "npx vite --config vite.cjs.config.ts build",
         "build:debug": "tsc --build --verbose",
         "build:test": "tsc --build --verbose tsconfig.test.json",
-        "build:config": "tsc --project",
+        "build:config": "tsc -project",
```
goastler: are you sure this is the cmd? my tsc help says:
```
 --project, -p  Compile the project given the path to its configuration file, o                 r to a folder with a 'tsconfig.json'.

```

tsconfig.cjs.json
```diff
@@ -2,6 +2,9 @@
     "extends": "./tsconfig.build.json",
     "compilerOptions": {
         "module": "CommonJS",
-        "target": "ES2015"
+        "target": "ES2020",
+        "resolveJsonModule": true,
+        "esModuleInterop": true,
+        "moduleResolution": "Node"
```
goastler, line 6: other than module resolution and module, these should be inherited from the tsconfig.build.json config

goastler (dismissed):

packages/provider/package.json
forgetso: yeah this is wrong

forgetso: search replace error

dev/config/src/dependencies.ts
forgetso: I've removed the map as the result is not used, on the recommendation of @withmaverick

goastler (approved):

PR #593, opened 2023-09-21
Adding dark theme option for CAPTCHA

HughParry:


packages/procaptcha-bundle/src/index.tsx
```diff
@@ -17,7 +17,6 @@ import { ProcapchaEventNames, ProcaptchaCallbacks, ProcaptchaConfigOptional } fr
 import { Procaptcha } from '@prosopo/procaptcha-react'
 import { at } from '@prosopo/util'
 import { createRoot } from 'react-dom/client'
-import React from 'react'
 
 function getConfig(siteKey?: string): ProcaptchaConfigOptional {
     if (!siteKey) {
```
ghost, line 22: Summary:
The code change involves the removal of an unused import statement, specifically `import React from 'react'`. This is a good practice as it helps to keep the codebase clean and efficient.

Feedback:
The change is positive as it removes an unnecessary import, which can potentially reduce the bundle size and improve the application's performance. It also enhances code readability and maintainability by removing clutter.

Suggestions:
1. Line 5: It's recommended to use absolute paths for imports to avoid potential issues with relative paths. This can be configured in the `tsconfig.json` or `jsconfig.json` file.
2. Line 29: Consider using a ternary operator for the `siteKey` assignment to make the code more concise.
3. Line 46: It's good practice to avoid using `any` type. Consider defining a type or an interface for the `window` object that includes the properties you're using.
4. Line 84: The console logs used for debugging should be removed in the production code to keep the console clean and improve performance.
5. Line 86: Consider using the `DOMContentLoaded` event listener in all cases, not just as a fallback. This ensures that the script runs after the entire document is loaded, which can prevent potential issues.
6. Line 90: The `ready` function is exported but not used anywhere else in the code. If it's not used in other modules, consider removing the export.

packages/procaptcha-bundle/src/index.tsx
```diff
@@ -107,19 +106,17 @@ export function render(callbacks?: ProcaptchaCallbacks) {
...
 export default function ready(fn: () => void) {
     if (document && document.readyState != 'loading') {
         console.log('document.readyState ready!')
         fn()
     } else {
         console.log('DOMContentLoaded listener!')
-        // note sure if this is the correct event listener
         document.addEventListener('DOMContentLoaded', fn)
     }
 }
```
ghost, line 122: Summary:
The code change is a refactor of the `render` function in a JavaScript module that deals with rendering a `Procaptcha` component. The `ready` function has also been slightly modified. The changes are minor and mostly involve cleaning up the code and removing unnecessary comments.

Feedback:
1. Line 106: The refactor of the `createRoot` function call is cleaner and more concise. It's a good practice to chain methods when possible to improve readability.
2. Line 107: The comment "wrap in fn and give user access to func" could be more descriptive. It's not clear what function is being wrapped or how the user is given access to it.
3. Line 116: The removal of the StackOverflow link is a good move as it's not a good practice to leave such comments in the code. If the link is important, it should be included in the documentation instead.
4. Line 119: The removal of the comment "note sure if this is the correct event listener" is also a good move. If there's uncertainty about the code, it should be resolved before committing, not noted in a comment.

Suggestions for improvement:
1. Line 107: Improve the comment to explain what function is being wrapped and how the user is given access to it.
2. Line 116 and 119: If the removed comments contained important information, consider adding that information to the documentation or resolving the uncertainty before committing.
3. General: Consider adding more comments to explain what the code does, especially for complex parts. This will make the code easier to understand for other developers.
4. General: Consider removing the console.log statements or replacing them with a proper logging system. Console.log statements are not suitable for production code.

forgetso (dismissed):

demos/client-bundle-example/src/index.html
```diff
@@ -1,24 +1,24 @@
...
+                <legend>Login</legend>
+                <div class="mui-textfield mui-textfield--float-label">
+                    <label>Email Address</label>
+                    <input type="email" required>
+                </div>
+                <div class="mui-textfield mui-textfield--float-label">
+                    <label>Password</label>
+                    <input type="password" required>
+                </div>
+                <div class="procaptcha" data-sitekey="5HUBceb4Du6dvMA9BiwN5VzUrzUsX9Zp7z7nSR2cC1TCv5jg" data-theme="dark"></div>
```
goastler, line 20: would you mind putting a comment saying this is the dev sitekey? just for inline docs

packages/procaptcha-react/src/components/CaptchaComponent.tsx
```diff
@@ -28,135 +28,143 @@ export interface CaptchaComponentProps {
...
-                        <Box
-                            px={2}
-                            py={3}
+                        <Typography
                             sx={{
-                                // center the header
-                                display: 'flex',
-                                alignItems: 'center',
-                                width: '100%',
+                                color: '#ffffff',
```
goastler: does this not need to be set based on the theme?

packages/procaptcha-react/src/components/CaptchaComponent.tsx
```diff
@@ -28,135 +28,143 @@ export interface CaptchaComponentProps {
...
+                            {': '}
+                        </Typography>
+                        <Typography
+                            px={1}
                             sx={{
-                                display: 'flex',
-                                alignItems: 'center',
-                                justifyContent: 'center',
-                                width: '100%',
+                                color: '#ffffff',
```
goastler: should this come from the theme?

packages/procaptcha-react/src/components/CaptchaComponent.tsx
```diff
@@ -28,135 +28,143 @@ export interface CaptchaComponentProps {
...
+                        {...addDataAttr({ dev: { cy: 'dots-captcha' } })}
+                    >
+                        {challenge.captchas.map((_, i) => (
+                            <Box
+                                key={i}
+                                sx={{
+                                    width: 7,
+                                    height: 7,
+                                    borderRadius: '50%',
+                                    border: '1px solid #CFCFCF',
```
goastler: should the colour come from the theme?

packages/procaptcha-react/src/components/CaptchaComponent.tsx
```diff
@@ -28,135 +28,143 @@ export interface CaptchaComponentProps {
...
+                                }}
+                                mx={0.5}
+                                bgcolor={index === i ? theme.palette.background.default : '#CFCFCF'}
```
goastler: should this come from the theme?

packages/procaptcha-react/src/components/CaptchaWidget.tsx
```diff
@@ -39,117 +38,113 @@ const getHash = (item: any) => {
...
-                                        src={item.data}
-                                        alt={`Captcha image ${index + 1}`}
-                                    />
-                                </Box>
-                                <Fade in={solution.includes(hash)}>
+                            <Box sx={{ border: 1, borderColor: 'lightgray' }}>
+                                <img
+                                    style={{
+                                        width: '100%', // image should be full width / height of the item
+                                        backgroundColor: 'lightgray', // colour of the bands when letterboxing and image
```
goastler: should this colour come from the theme?

goastler (changes requested):
Few questions regarding colour, as I think these need to be derived from the theme otherwise clashes could happen

packages/procaptcha-react/src/components/CaptchaComponent.tsx
HughParry: No, this is the topbar for the expanded captcha which is blue background white text

packages/procaptcha-react/src/components/CaptchaComponent.tsx
HughParry: No, this is the topbar for the expanded captcha which is blue background white text

packages/procaptcha-react/src/components/CaptchaComponent.tsx
HughParry: Could do - What colour do you want to go for the image border for dark mode?

Personally I'm not completely convinced that the images really need borders, so happy with whatever you want to go for on this one

packages/procaptcha-react/src/components/CaptchaWidget.tsx
HughParry: Potentially - this depends on how we want to handle theming in the future, so I'd prefer to hold off moving the non changing styling into the MUI theme if that's not the way we end up handling the style customisability long term

goastler (dismissed):
lgtm

goastler (approved):

PR #597, opened 2023-09-21
make contract and crate version the same (0.2.2)

goastler:


goastler:
fixes prosopo/captcha-private#520

forgetso (approved):

PR #602, opened 2023-09-22
workflow to check if contract and crate versions match provider version

goastler:


goastler:
fixes prosopo/captcha-private#519

forgetso (approved):
nice

PR #606, opened 2023-09-22
Eslint regexp

goastler:


goastler:
fixes prosopo/captcha-private#517

goastler:
@forgetso could you have a close look at the regex's to make sure they're working as you intended? the linter changed a couple of them :)

forgetso (approved):
nice

PR #620, opened 2023-09-24
Reusable version check workflow

goastler:


goastler:
This is the reusable workflow, so the one that can be called for other workflows. Once it's in main I'll switch over other workflows to call it directly rather than having the version checking duplicated amongst ~4 workflows

forgetso (approved):

PR #624, opened 2023-09-25
Issue 596/replicate hcaptcha recaptcha callback naming system

HughParry:


demos/client-bundle-example/src/index.html
```diff
@@ -1,25 +1,43 @@
 <html>
     <head>
-        <link href="//cdn.muicss.com/mui-0.10.3/css/mui.min.css" rel="stylesheet" type="text/css"/>
+        <link href="//cdn.muicss.com/mui-0.10.3/css/mui.min.css" rel="stylesheet" type="text/css" />
         <title>Procaptcha demo: Simple page</title>
-        <script src="procaptcha.bundle.js" async defer></script>
+        <script src="procaptcha.bundle.js"></script>
```
forgetso: Is there a reason why `async defer` was removed? It's in both [recaptcha's docs](https://developers.google.com/recaptcha/docs/display#auto_render) and [hcaptcha's docs](https://docs.hcaptcha.com/#add-the-hcaptcha-widget-to-your-webpage). Otherwise, looks good!

forgetso (changes requested):

demos/client-bundle-example/src/index.html
HughParry: Result of testing with script tags in different places and not including an event listener for DOMContentLoaded - will add back in

forgetso (approved):

PR #625, opened 2023-09-25
622 update readmes

forgetso:


README.md
```diff
@@ -17,7 +17,7 @@ placed anywhere on the page. Inside the <head> tag or immediately after the .pro
```
HughParry, line 19: We can also include render method, with similar docs to https://docs.hcaptcha.com/configuration#explicitly-render-hcaptcha

packages/procaptcha-react/README.md
```diff
@@ -1,8 +1,9 @@
-# React Prosopo Procaptcha Component Library
+# Prosopo Procaptcha React Component Library
 
 React components for integrating the Prosopo [procaptcha](https://github.com/prosopo/procaptcha) into a React app.
 
-Prosopo is a distributed human verification service that can be used to stop bots from interacting with smart contracts. Sign up to be a network [alpha tester](https://5b06hrhtlmh.typeform.com/to/vNpyOUfg).
+Prosopo is a distributed human verification service that can be used to stop bots from interacting with smart contracts.
+Sign up to be a network [alpha tester](https://5b06hrhtlmh.typeform.com/to/vNpyOUfg).
```
HughParry: Beta test not alpha, and should link to prosopo sign up form not the typeform

packages/cli/README.md
```diff
@@ -0,0 +1,72 @@
+# Prosopo Provider Command Line Interface
```
HughParry, line 1: Unless this is just for internal use then this should have a little more context imo

What does fund do, why should I do it, how much should I do it, how do i get dataset.json, (etc that kind of context)

HughParry (dismissed):

HughParry (approved):

PR #627, opened 2023-09-26
Procaptcha tech debt fixes

HughParry:


packages/procaptcha/package.json
```diff
@@ -7,9 +7,9 @@
     "type": "module",
     "types": "./dist/index.d.ts",
     "scripts": {
-        "test": "NODE_ENV=test ts-mocha \"tests/**/*.test.ts\" && exit 0",
         "clean": "tsc --build --clean",
-        "build": "tsc --build --verbose tsconfig.json",
+        "prebuild": "node prebuild-write-version.js",
```
goastler, line 15: I __think__ this conflicts with https://docs.npmjs.com/cli/v10/using-npm/scripts#npm-run-user-defined

You can probably make use of the above instead, i.e. don't worry about calling the prebuild script yourself because npm should do it for you

goastler (changes requested):

packages/procaptcha/src/api/ExtensionWeb2.ts
```diff
@@ -76,7 +77,7 @@ export default class ExtWeb2 extends Extension {
                 },
             },
             name: 'procaptcha-web2',
-            version: '0.1.11',
+            version,
```
forgetso, line 81: Would this not be better / simpler as a bundler step? e.g. `process.env.PROCAPTCHA_VERSION`

packages/procaptcha/package.json
HughParry: yeah this works

forgetso (changes requested):

PR #652, opened 2023-09-29
Implement config defaults

forgetso:


forgetso:
- [ ] TODO: Obtain SS58_FORMAT and PAIR_TYPE from polkadot package

demos/client-example/src/config.ts
```diff
@@ -17,37 +17,30 @@
...
+import { EnvironmentTypesSchema, NetworkNamesSchema } from '@prosopo/types'
+import { ProsopoClientConfigInput } from '@prosopo/types'
 
-const { REACT_APP_PROSOPO_CONTRACT_ADDRESS, REACT_APP_SUBSTRATE_ENDPOINT } = z
-    .object({
-        REACT_APP_PROSOPO_CONTRACT_ADDRESS: z.string(),
-        REACT_APP_SUBSTRATE_ENDPOINT: z.string(),
-    })
-    .parse(process.env)
+const getWeb2 = (): boolean | undefined => {
```
HughParry, line 23: Could be much shorter snippet. If we default to true (which is potentailly expected behaviour) then:

return !(process.env.REACT_APP_WEB2 === 'false')

---

if we require undefined then something like:

process.env.REACT_APP_WEB2 === 'true' ? true : process.env.REACT_APP_WEB2 === 'false' ? false : undefined;

dev/scripts/src/scripts/getNetworks.ts
```diff
@@ -0,0 +1,38 @@
...
+import { selectableNetworks } from '@polkadot/networks'
+import { testParasRococoCommon } from '@polkadot/apps-config'
+async function run() {
+    // // Construct
+    // const provider = 'ws://localhost:9944'
+    // // const provider = 'wss://shiden.public.blastapi.io'
+    // const wsProvider = new WsProvider(provider)
+    // const api = await ApiPromise.create({ provider: wsProvider })
+
+    selectableNetworks.forEach((network) => {
```
HughParry, line 23: Can we be consistent on using functional methods vs for loops within files(sorry this one pretty nitpicky)

HughParry (changes requested):

dev/scripts/src/scripts/getNetworks.ts
forgetso: This wasn't meant to be in the PR. I've deleted it.

demos/client-example/src/config.ts
forgetso: Changed

HughParry (approved):

PR #683, opened 2023-10-10
Dataset fs cli restructure

goastler:
- Fixes bugs with resizing images and flattening datasets
- Restructures the cli into a simpler format


packages/datasets-fs/src/commands/flatten.ts
```diff
@@ -0,0 +1,122 @@
...
+            0
+        )
+        for (let i = 0; i < labels.length; i++) {
+            // find all the images
+            const label = at(labels, i)
+            const images: string[] = at(imagesByLabel, i)
+            // for each image
+            for (const image of images) {
+                bar.increment()
+                // this.logger.log(`flattening ${label}/${image}`)
```
HughParry, line 73: Should this be commented out?

packages/datasets-fs/src/commands/generate.ts
```diff
@@ -0,0 +1,169 @@
...
+    labelled: LabelledItem[],
+    unlabelled: Item[],
+    options: {
+        allowDuplicatesLabelled?: boolean
+        allowDuplicatesUnlabelled?: boolean
+    }
+) => {
+    // check for duplicates
+    const all = new Set<string>()
+    if (!options.allowDuplicatesLabelled) {
```
HughParry: Error handling here could be shared function, with only all.add changing depending on labelled/unlabelled

packages/datasets-fs/src/commands/generate.ts
```diff
@@ -0,0 +1,169 @@
...
+    labelled: LabelledItem[] = []
+    unlabelled: Item[] = []
+    labels: string[] = []
+    labelledMapFile = ''
+    unlabelledMapFile = ''
+    labelToImages: { [label: string]: Item[] } = {}
+    targets: string[] = []
+    saltRounds = 10
+
+    public override async _run(args: Args) {
```
HughParry: Pretty lengthy in places - could we break some parts of _run down into helper funcs like done with checkduplicated already on line 115?

packages/datasets-fs/src/commands/generateV1.ts
```diff
@@ -0,0 +1,207 @@
...
+                number: true,
+            },
+            maxCorrect: {
+                description: 'Maximum number of target images in each captcha',
+                number: true,
+            },
+        })
+    }
+
+    public override async _run(args: Args) {
```
HughParry: helper functions would be useful for readability here - _run is about 150 lines long.

Might not be worth doing if we aren't looking to be using v1 captcha generation in the future though

packages/datasets-fs/src/commands/generateV2.ts
```diff
@@ -0,0 +1,204 @@
...
+                description: 'Minimum number of labelled images in each captcha',
+            },
+            maxLabelled: {
+                number: true,
+                description: 'Maximum number of labelled images in each captcha',
+            },
+        })
+    }
+
+    public override async _run(args: Args) {
```
HughParry, line 122: _run is a long method here, will be hard to maintain. Might be worth breaking up a little

packages/datasets-fs/src/commands/generateV2.ts
```diff
@@ -0,0 +1,204 @@
...
+        // if we focus on a single captcha round of 9 images, we must have at least 1 labelled correct image in the captcha for it to work, otherwise it's just a labelling phase, which normally isn't a problem but if we're treating these as tests for humanity too then we need some kind of test in there. (e.g. we abolish the labelled then unlabelled pattern of the challenge rounds in favour of mixing labelled and unlabelled data, but we then run a small chance of serving two completely unlabelled rounds if we don't set the min number of labelled images to 1 per captcha round)
+
+        const bar = new cliProgress.SingleBar({}, cliProgress.Presets.shades_classic)
+        bar.start(count, 0)
+
+        // generate n captchas
+        const captchas: CaptchaWithoutId[] = []
+        for (let i = 0; i < count; i++) {
+            bar.increment()
+            // this.logger.info(`generating captcha ${i + 1} of ${count}`)
```
HughParry, line 152: Should be able to remove this comment with the bar implementation

packages/datasets-fs/src/commands/generateV2.ts
```diff
@@ -0,0 +1,204 @@
...
+
+            // get the unlabelled items
+            const unlabelledItems = new Set<Item>()
+            while (unlabelledItems.size < size - nLabelled) {
+                // get a random image from the unlabelled data
+                const image = at(this.unlabelled, _.random(0, this.unlabelled.length - 1))
+                unlabelledItems.add(image)
+            }
+
+            let items: Item[] = [...correctItems, ...incorrectItems, ...unlabelledItems]
```
HughParry: Could we use functional methods here to avoid use of let

packages/datasets-fs/src/commands/resize.ts
```diff
@@ -0,0 +1,137 @@
...
+        // read the map file
+        const inputItems: Item[] = DataSchema.parse(JSON.parse(fs.readFileSync(mapFile, 'utf8'))).items
+
+        // for each item
+        const bar = new cliProgress.SingleBar({}, cliProgress.Presets.shades_classic)
+        bar.start(inputItems.length, 0)
+        const outputItems: Item[] = []
+        for (const inputItem of inputItems) {
+            bar.increment()
+            // this.logger.log(`resizing ${inputItem.data}`)
```
HughParry, line 75: Remove logger comment

packages/datasets-fs/src/tests/utils.ts
```diff
@@ -0,0 +1,132 @@
...
+            return false
+        }
+    }
+    return true
+}
+
+export const readDataJson = (pth: string) => {
+    let content = fs.readFileSync(pth).toString()
+    // TODO use getPaths() here to find the repo dir
+    content = content.replaceAll('${repo}', __dirname + '/../../../..')
```
HughParry, line 66: Is there a ticket for the above TODO - magic strings for /../../../.. are fine for now but getPaths() would be better

packages/datasets-fs/src/tests/utils.ts
```diff
@@ -0,0 +1,132 @@
...
+    const dataJson = JSON.parse(content.toString())
+    const data = DataSchema.parse(dataJson)
+    return data
+}
+
+export const readCaptchasJson = (pth: string) => {
+    const content = fs.readFileSync(pth).toString()
+    const dataJson = JSON.parse(content.toString())
```
HughParry: Calling toString() twice on content

packages/util/src/util.ts
```diff
@@ -197,3 +200,109 @@ export const flattenObj = (obj: object, prefix = ''): Record<string, unknown> =>
...
+// This method treats arrays as an object with numeric keys and merged using the object merge strategy.
+export function merge<T extends object | A[], U extends object | B[], A, B>(
+    dest: T,
+    src: U,
+    options?: MergeOptions
+): T & U {
+    const atomicArrays = options?.atomicArrays
+    // maintain a queue of object sources/destinations to merge
+    const queue: {
+        src: unknown
```
HughParry, line 229: Can we infer types of src and dest or is it better to leave as unknown?

HughParry (changes requested):

packages/datasets-fs/src/commands/flatten.ts
goastler: yes because it messes with the progress bar that chris added. Unfortunately the progress bar lib cannot deal with printlns when it's active

packages/datasets-fs/src/commands/resize.ts
goastler: same deal as above

packages/datasets-fs/src/tests/utils.ts
goastler: there isn't, but getPaths already exists. It's just in the config package and I didn't want to introduce another dependency. @forgetso can we depend on the config pkg or will that introduce circular deps?

packages/datasets-fs/src/commands/generate.ts
goastler: done

packages/datasets-fs/src/commands/generate.ts
goastler: done

packages/datasets-fs/src/commands/flatten.ts
HughParry: Can we remove instead of just commenting out?

packages/datasets-fs/src/tests/utils.ts
forgetso: The config package is really only a dev dependency. You'd need to move the path finding stuff into a new package and make it a regular dependency.

packages/datasets-fs/src/commands/generateV2.ts
goastler: done

packages/datasets-fs/src/tests/utils.ts
goastler: done

packages/util/src/util.ts
goastler: it's better to leave these as unknown as we don't know what type we're handling. Usually it'll be an array or an object, but it may be a primitive if the user passes in parameters for that. I did have types in here but the bounds became super complex, so I switched to unknown because this fn should handle anything and everything

HughParry (dismissed):

HughParry (approved):

PR #686, opened 2023-10-10
Public account for prosopo server

HughParry:


packages/server/src/index.ts
```diff
@@ -146,3 +147,9 @@ export class ProsopoServer {
         return this.contract
     }
 }
+
+export const PublicProsopoServer = async (config: ProsopoServerConfig, publicAccountNetwork: PublicAccountNetwork) => {
+    const publicProsopoPai = await getPublicProsopoPair(publicAccountNetwork)
```
goastler: typo "pai"

packages/server/src/publicProsopoPair.ts
```diff
@@ -0,0 +1,16 @@
...
+export type PublicAccountNetwork = 'rococo'
+
+export const getPublicProsopoPair = async (network: PublicAccountNetwork) => {
+    const accountNetworkMap = {
+        rococo: 'open poet tennis test nice immune scorpion flag cement surge effort canal', // Prosopo rococo shared public account
+    }
+
+    const pairType = 'sr25519' as KeypairType
+    const ss58Format = 42
+    const prosopoSharedAccountSecret = accountNetworkMap[network]
```
goastler, line 13: `prosopoSharedAccountSecret` must be `string | undefined`, no? Think you need to use the `get` fn from utils

goastler (changes requested):

packages/server/src/publicProsopoPair.ts
HughParry: Just comes up as string in my IDE - typescript actually being clever for once?

goastler: ah no, my bad. It's because you've set the type to `PublicAccountNetwork` rather than string. All good :)

goastler (approved):

PR #694, opened 2023-10-16
Fail to start API if env.db is undefined

HughParry:


goastler (dismissed):

packages/cli/src/process.env.ts
```diff
@@ -44,3 +44,7 @@ export function getSecret(who?: string): string {
 export function getConfig(): ProsopoConfig {
     return prosopoConfig() as ProsopoConfig
 }
+
+export function getDB(): string {
+    return process.env.db || ''
```
goastler: can you put the throw here? It's simpler imo to have a getter which always returns a valid value and throws otherwise. Saves us from doing the check everywhere the getter is used, also

goastler (changes requested):

goastler (dismissed):

goastler (approved):

PR #731, opened 2023-10-24
add cargo target

goastler:


forgetso (approved):

goastler:
this doesn't work with cargo contract it seems, ending up with a broken setup

PR #748, opened 2023-10-26
747 reduce bundle size by dropping unnecessary polkadotjs interfaces and upgrade arrays

forgetso:
- [x] Bundle size from 1.7MB to 1.2MB
- [x] Lodash moved to named export in @prosopo/util
- [x] zod exports changed to be tree shakeable in future
- [x] @polkadot/api imports changed to be tree shakeable
- [x] @prosopo/captcha-contract named imports and imports changed to be tree shakeable
- [x] Loads of interfaces, genesis info, api derive stuff from polkadot/api aliased to be empty files to save on space
- [x] environment variables fixed in env.rococo and env.test templates
- [x] speedier linting and prettier command created for building and linting contracts
- [x] custom wasm bytes object created using https://github.com/prosopo/wasm/tree/slim


packages/contract/src/contract/index.ts
```diff
@@ -11,7 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-import '@polkadot/api-augment'
+// import '@polkadot/api-augment'
```
HughParry: Why commented out?

forgetso: Because it only needs to be imported once and it's imported in `packages/contract/src/index.ts`. I was experimenting with removing it but it breaks the TS build. I've deleted the commented one.

HughParry (approved):

PR #756, opened 2023-10-31
drop http/https prefix from captcha data urls

goastler:


packages/procaptcha/src/modules/ProsopoCaptchaApi.ts
```diff
@@ -60,6 +60,21 @@ export class ProsopoCaptchaApi {
         try {
             const captchaChallenge = await this.providerApi.getCaptchaChallenge(this.userAccount, this.provider)
             this.verifyCaptchaChallengeContent(this.provider, captchaChallenge)
+            // convert https/http to match page
+            captchaChallenge.captchas.forEach((captcha) => {
+                captcha.captcha.items.forEach((item) => {
+                    if (item.data) {
+                        // drop the 'http(s):' prefix, leaving '//'. The '//' will autodetect http/https from the page load type
+                        // https://stackoverflow.com/a/18320348/7215926
+                        if (item.data.startsWith('https://')) {
```
forgetso: Any reason not to use a regex?

```
item.data = item.data.replace(/^http(s)*:\/\//, '')
```

And should we be supporting `file://` [et al.](https://en.wikipedia.org/wiki/List_of_URI_schemes)?

forgetso (changes requested):

packages/procaptcha/src/modules/ProsopoCaptchaApi.ts
goastler: regex might replace any http(s) string in another part of the url, e.g. a hash, e.g. 123abc1http23 becomes 123abc123

file:// still works with this mechanism, it leaves it untouched

forgetso: But the hat at the beginning of the regex command makes sure it's the start of the string. `^h`

goastler: good point, I hadn't spotted the hat! will update :)

regarding the `file://`, I don't believe that triggers a mixed-content warning as the warning is only concerned about using http/https requests when the page is loaded in the other manner. `file://` is local and has no comparison to http(s) because it isn't encrypted whatsoever, doesn't need to be, so it _shouldnt_ trigger the warning. But I'll check :)

goastler: only affects http(s), not other protocols :)

forgetso (approved):

PR #897, opened 2023-12-08
Drop mui

forgetso:
Remove the mui package and create custom components to save space in the bundle.


packages/procaptcha-react/src/components/CaptchaWidget.tsx
```diff
@@ -91,58 +83,70 @@ export const CaptchaWidget = ({ challenge, solution, onClick }: CaptchaWidgetPro
...
+                                style={{
+                                    // relative to where the element _should_ be positioned
+                                    position: 'relative',
+                                    // make the overlay the full height/width of an item
+                                    width: '100%',
+                                    height: '100%',
+                                    // shift it up 100% to overlay the item element
+                                    top: '-100%',
+                                    // transition on opacity upon (de)selection
+                                    //transitionDuration: '300ms',
```
HughParry: Can we remove these unused commented out properties?

In general there's a few comments in this bit that could easily become stale and confusing

packages/procaptcha-react/src/components/ExtensionAccountSelect.tsx
```diff
@@ -44,25 +40,27 @@ export const ExtensionAccountSelect = ({
...
                 } else {
                     console.log('Deselected account')
                     onChange('')
                 }
             }}
-            sx={{ width: 550 }}
-            getOptionLabel={(option: InjectedAccountWithMeta) => `${option.meta.name}: ${option.address}`}
-            renderInput={(props) => <TextField {...props} label={t('WIDGET.SELECT_ACCOUNT')} />}
-        />
+            value={accounts.length > 0 && account ? account.toString() : undefined}
```
HughParry: should this be account.address not account.toString()?

packages/procaptcha-react/src/components/Modal.tsx
```diff
@@ -0,0 +1,50 @@
...
+            transition: 'all 0.5s',
+        }
+
+        return (
+            <div style={ModalOuterDivCss}>
+                <div style={ModalInnerDivCSS}>{children}</div>
+            </div>
+        )
+    }
+    // (prevProps, nextProps): boolean => {
```
HughParry: Can we remove this please?

packages/procaptcha-react/src/components/LoadingSpinner.tsx
```diff
@@ -0,0 +1,35 @@
...
+/** @jsxImportSource @emotion/react */
+import { darkTheme, lightTheme } from './theme.js'
+import { useMemo } from 'react'
+import styled from '@emotion/styled'
+
+export interface LoadingSpinnerComponentProps {
+    themeColor: 'light' | 'dark'
+}
+
+export const LoadingSpinner = ({ themeColor }: LoadingSpinnerComponentProps) => {
```
HughParry, line 11: Should have aria labels on stuff like this. Not super important right now since we don't have accessibility options impl right now, but at the least is SEO friendly for the dapps hosting us

packages/procaptcha-react/src/components/Modal.tsx
HughParry: Assuming that the default memo behaviour is comparing props and nextprops correctly

packages/procaptcha-react/src/components/Procaptcha.tsx
```diff
@@ -205,52 +217,44 @@ export const Procaptcha = (props: ProcaptchaProps) => {
...
+                                        </div>
+                                    </div>
+                                </div>
+                                <div style={{ padding: 1 }}>
+                                    <span style={{ color: theme.palette.primary.contrastText }}>I am a human</span>
+                                </div>
+                            </div>
+                        </div>
+                        <div>
+                            <a href="https://prosopo.io" target="_blank">
```
HughParry: It could be smart to change this to a FAQ for Procaptcha, then set rel="help"

packages/procaptcha-react/src/components/theme.ts
```diff
@@ -11,30 +11,46 @@
...
+    900: '#212121',
+}
+
+export const lightTheme = {
     palette: {
         mode: 'light',
         primary: {
-            main: '#1976d2',
-            contrastText: '#000',
+            main: '#542437',
```
HughParry: Are these equiv?

HughParry (changes requested):

packages/procaptcha-react/src/components/ExtensionAccountSelect.tsx
forgetso: yes, good point!

packages/procaptcha-react/src/components/CaptchaWidget.tsx
forgetso: :+1:

packages/procaptcha-react/src/components/Modal.tsx
forgetso: Seems like it is.

packages/procaptcha-react/src/components/LoadingSpinner.tsx
forgetso: Sure, will add them.

packages/procaptcha-react/src/components/Procaptcha.tsx
forgetso: Very good point.

packages/procaptcha-react/src/components/theme.ts
forgetso: No, I changed the colour to something other than blue. maybe I shouldn't have. Will set a poll in slack.

packages/procaptcha-react/src/components/LoadingSpinner.tsx
forgetso: I've added the aria labels to the checkbox.

packages/procaptcha-react/src/components/Procaptcha.tsx
forgetso: Have change it to https://www.prosopo.io/#features for now

HughParry (approved):

PR #930, opened 2024-01-08
Prosopo error simplification

HughParry:


demos/provider-mock/src/api.ts
```diff
@@ -35,9 +35,12 @@ export function prosopoRouter(): Router {
         try {
             parsed = VerifySolutionBody.parse(req.body)
         } catch (err) {
-            // TODO fix error handling
-            console.log('parsing error')
-            return next(new ProsopoApiError(err as Error, undefined, 400))
+            return next(
+                new ProsopoApiError('CAPTCHA.PARSE_ERROR', {
+                    context: { error: err, errorCode: 400 },
+                    logLevel: 'info',
```
goastler, line 41: so if I don't specify a logLevel, the error will not print, correct?

packages/common/src/error.ts
```diff
@@ -11,93 +11,147 @@
...
+
+    private logError(logger: Logger, logLevel: LogLevel) {
+        const errorFormatter = '\n********************* ERROR *********************\n'
+        const errorName = `Error Type: ${this.name}\n`
+        const errorParams = JSON.stringify({ error: this.message, context: this.context })
+        const errorMessage = `${errorFormatter}${errorName}${errorParams}`
+
+        switch (logLevel) {
+            case 'trace':
+                logger.trace(errorMessage, this.stack)
```
goastler: is there not a method to do something like `logger.log(level, msg)`?

packages/common/src/error.ts
```diff
@@ -11,93 +11,147 @@
...
-        console.error('\n********************* ERROR *********************\n')
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-ignore
-        if (typeof this.cause != ZodError && this.cause?.message && this.cause.message.length > MAX_ERROR_LENGTH) {
-            this.cause.message = `${at(this.cause.message, -MAX_ERROR_LENGTH)}...`
+        this.logError(logger, logLevel)
+    }
+
+    private logError(logger: Logger, logLevel: LogLevel) {
+        const errorFormatter = '\n********************* ERROR *********************\n'
```
goastler: can we reduce the width of this? always wraps in my console

packages/common/src/error.ts
```diff
@@ -11,93 +11,147 @@
...
+type DatasetErrorOptions = ErrorOptions & { context?: DatasetContextParams }
+
+type ApiContextParams = ContextParams & { errorCode?: number }
+type ApiErrorOptions = ErrorOptions & { context?: ApiContextParams }
+
 export abstract class ProsopoBaseError extends Error {
-    protected tKey: TranslationKey | undefined
-    protected tParams: Record<string, string> | undefined
+    translationKey: string | undefined
+    context: ContextParams | undefined
```
goastler: try making ProsopoBaseError generic, e.g. `class ProsopoBaseError<Context>` then in the class have the context fields as `context: Context | undefined`, then classes which extends this error can set their context type like so: `class ProsopoContractError extends ProsopoBaseError<ContractErrorOptions>` ...

if this doesn't make sense I can create a quick mockup

goastler (changes requested):

demos/client-example-server/src/app.ts
```diff
@@ -55,8 +57,13 @@ async function main() {
     const uri = await memoryServerSetup()
     console.log('mongo uri', uri)
     const mongoose = connectionFactory(uri)
-    if (!process.env.PROSOPO_SITE_PRIVATE_KEY) {
-        throw new Error('No private key found')
+    if (!process.env.REACT_APP_SERVER_MNEMONIC) {
```
forgetso: What's the reason for the env variable name? All env vars should now start with PROSOPO

packages/common/src/utils.ts
```diff
@@ -43,7 +44,7 @@ function getLeafFieldPath(obj: Node): string[] {
     return Object.keys(obj).reduce((arr, key) => {
         const value = obj[key]
         if (value === undefined) {
-            throw new Error(`Undefined value for key ${key}`)
+            throw new ProsopoError('DEVELOPER.GENERAL', { context: { error: `Undefined value for key ${key}` } })
```
forgetso: Create a `KeyError` type

packages/datasets/src/captcha/merkle.ts
```diff
@@ -94,7 +94,9 @@ export class CaptchaMerkleTree {
         while (layerNum < this.layers.length - 1) {
             const layer = this.layers[layerNum]
             if (layer === undefined) {
-                throw new Error('layer undefined')
+                throw new ProsopoError('DEVELOPER.GENERAL', {
```
forgetso: `MerkleError`

forgetso (changes requested):
Really tidies things up. A few small changes required.

demos/provider-mock/src/api.ts
HughParry: Log level determines the log level of the output message. It defaults to error.

If a logger with a log level set to, e.g. error, is passed in, then only errors with logLevel: "error" or above will actually print.

e.g.

const logger = new Logger(logLevel: "error")
new ProsopoError(error_code, {logLevel: 'info', logger}) <-- This will not print anything
new ProsopoError(error_code, {logLevel: 'fatal', logger}) <-- This prints as fatal
new ProsopoError(error_code, {logger}) <-- This prints as error

packages/common/src/error.ts
HughParry: @forgetso to resolve by purchasing a larger monitor ;)


@goastler Does it wrap at this reduced length?

packages/common/src/error.ts
HughParry: We'd have to impl this in the logger.ts file, which would effectively just be moving the switch statement there. Didn't want to do this as I was concerned it would mess with previous usage of logger.log

packages/common/src/utils.ts
HughParry: Done

packages/datasets/src/captcha/merkle.ts
HughParry: Done

packages/common/src/error.ts
HughParry: Done

demos/client-example-server/src/app.ts
HughParry: I assume comparing and making changes vs an out of date file. Should be fixed now

demos/provider-mock/src/api.ts
goastler: what happens in the case where I've got an error that I don't want printed because I'm going to handle it? e.g. I call a fn which fails due to a failed fetch request. I want to simply retry calling this fn again. The error shouldn't be printed whatsoever

packages/common/src/error.ts
goastler: https://github.com/prosopo/captcha/pull/983

packages/common/src/error.ts
goastler: yeah anything shorter will be fine

packages/common/src/error.ts
HughParry: clean fix nice

HughParry: addressed in latest push

demos/provider-mock/src/api.ts
HughParry: then either

1. pass your logger and a log level lower than that logger
2. pass the log level as trace - the default log level is info, so trace won't print anything

goastler: sure, so proceeding with option 2, if I've got my logger set to trace then it would still print. Do you see my point? there's always a configuration which results in the error being printed. imo we should have the ability not to print anything at all, as some errors should be handled silently

forgetso (dismissed):

demos/provider-mock/src/api.ts
HughParry: sure 2 secs I'll add a silent config option then

goastler (dismissed):

forgetso (approved):

PR #984, opened 2024-01-15
Provider gui config from packages

HughParry:


provider-gui/services/contract/config.ts
```diff
@@ -0,0 +1,49 @@
...
+        defaultEnvironment: environment,
+        defaultNetwork: network,
+        account: {
+            password: '',
+            address: 'currentAccount',
+        },
+        database: {
+            development: { type: DatabaseTypes.enum.mongoMemory, endpoint: '', dbname: 'prosopo_test', authSource: '' },
+        },
+        server: {
```
forgetso: This is for the image server. You shouldn't need to declare it.

provider-gui/services/contract/config.ts
```diff
@@ -0,0 +1,49 @@
...
+            baseURL: 'http://localhost',
+            port: 9229,
+            fileServePaths: '[]',
+        },
+        networks: {
```
forgetso: This should default to sensible values if you don't supply it.

https://github.com/prosopo/captcha/blob/d5b42d16c2a10aec21a8d8b57a211a1f8e0dde07/packages/types/src/networks/index.ts#L10-L40

forgetso (changes requested):
Couple of very minor points.

provider-gui/services/contract/config.ts
HughParry: Removed

provider-gui/services/contract/config.ts
HughParry: Sorted

forgetso (approved):

PR #996, opened 2024-01-18
Provider admin api with signing

HughParry:


packages/datasets/src/tests/mocks/data/captchas.json
```diff
@@ -2,7 +2,7 @@
     "datasetId": "0x80f2e82ed079330c5fd9ea849a2b068e5b5c3473061219e5b33f2f210e6f1715",
     "captchas": [
         {
-            "salt": "0x01",
+            "salt": "0x0ee5e8130bf5834978971abbb3e94703",
```
forgetso, line 5: Why have these salts changed?

forgetso (changes requested):
On track for world domination. Good work. One small Q. We can handle the commit stuff later when the populator is working.

packages/datasets/src/tests/mocks/data/captchas.json
HughParry: IIRC, zod was getting angry about arg length for salts being under 34 chars

forgetso: Great

forgetso (approved):

PR #998, opened 2024-01-24
Make admin API optional

forgetso:


HughParry (commented):
Was the bug that non admin endpoints were still asking for auth?

goastler (dismissed):

forgetso:
> Was the bug that non admin endpoints were still asking for auth?

Yeah, switching on the api automatically assumed that a token would be present for the admin API. It's cause of the way it was added to the express router. It's initialised with `apiApp.use(prosopoAdminRouter(env))` so we need some kind of flag to stop this from happening. Open to better suggestions.

goastler (approved):

PR #999, opened 2024-01-24
Test for reloading API

forgetso:


packages/cli/src/tests/reloader.test.ts
```diff
@@ -0,0 +1,81 @@
+import { describe, expect, test } from 'vitest'
+import { getCurrentFileDirectory } from '@prosopo/util'
+import { promisify } from 'util'
+import { spawn } from 'child_process'
+import fs from 'fs'
+import path from 'path'
+
+describe('reloading api', () => {
+    test.only('api reloads after changing .env file', async () => {
```
goastler: what's the reason you're using `only` here?

goastler (changes requested):

packages/cli/src/tests/reloader.test.ts
forgetso: it's a mistake.

goastler (approved):

PR #1001, opened 2024-01-25
Chunked bundling and lazy loading of components

HughParry:


forgetso (approved):
Remember to check the test logs!

PR #1010, opened 2024-01-31
Fix vite/vitest hanging

goastler:
fix i18n setup to avoid setInterval call, allowing node to exit cleanly


forgetso (approved):
Well done

PR #1026, opened 2024-02-07
Fix docker js_server image

forgetso:
The image creation was broken since the bundle is now split into multiple parts.


forgetso:
Closing in favour of https://github.com/prosopo/captcha/pull/1027

PR #1039, opened 2024-02-20
Pow v1 impl

HughParry:


demos/client-bundle-example/src/index.html
```diff
@@ -22,6 +22,7 @@
                 <div
                     class="procaptcha"
                     data-theme="light"
+                    data-captcha-type="image"
```
goastler: still set to image captcha?

packages/procaptcha-bundle/src/index.tsx
```diff
@@ -20,12 +20,16 @@ import {
     ProcaptchaOutput,
 } from '@prosopo/types'
 import { Procaptcha } from '@prosopo/procaptcha-react'
+import { ProcaptchaPow } from '@prosopo/procaptcha-pow'
 import { at } from '@prosopo/util'
 import { createRoot } from 'react-dom/client'
 
+const features = ['image', 'pow'] as const
```
goastler: this needs to be moved to elsewhere outside of procaptcha-bundle

goastler (dismissed):
looks good other than a lot of duplication, which I think needs to be reduced before merging in. I don't mind some duplication where inconvenient to deduplicate

demos/client-bundle-example/src/index.html
forgetso: do we need a new captcha type in the contract?

demos/client-pow-example/env.development
```diff
@@ -0,0 +1,8 @@
+PROSOPO_SITE_KEY=5HGjWAeFDfFCWPsjFQdVV2Msvz2XtMktvgocEZcCj68kUMaw
+PROSOPO_SUBSTRATE_ENDPOINT=ws://127.0.0.1:9944
+PROSOPO_CONTRACT_ADDRESS=
+PROSOPO_WEB2=true
+PROSOPO_SERVER_URL=http://localhost:9228
+PROSOPO_PORT=9230
```
forgetso: Please use a different port as 9230 is used in the client-example demo.

demos/client-pow-example/vite.config.ts
```diff
@@ -0,0 +1,60 @@
...
+        },
+        plugins: [
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore
+            react(),
+            // Closes the bundler and copies the bundle to the client-bundle-example project unless we're in serve
+            // mode, in which case we don't want to close the bundler because it will close the server
+            command !== 'serve' ? VitePluginCloseAndCopy() : undefined,
+        ],
+        server: { port: process.env.PROSOPO_PORT ? Number(process.env.PROSOPO_PORT) : 9230 },
```
forgetso: Use a different default port.

demos/client-pow-example/vite.config.ts
```diff
@@ -0,0 +1,60 @@
+import * as path from 'path'
```
forgetso, line 1: It might be good to bring this config into the `config` package and use as a base for all vite frontend configs. Later though.

packages/procaptcha-pow/src/Services/canvas.ts
```diff
@@ -0,0 +1,435 @@
+// Copyright 2021-2023 Prosopo (UK) Ltd.
```
forgetso: Is this file exactly the same as canvas in the other package? If so, stick it in its own package or in util or something.

packages/provider/src/tasks/tasks.ts
```diff
@@ -164,6 +167,138 @@ export class Tasks {
         })
     }
 
+    /**
+     * @description Generates a PoW Captcha for a given user and dapp
+     *
+     * @param {string} userAccount - Dapp User address
+     * @param {string} dappAccount - Dapp User address
+     */
+    async getPowCaptchaChallenge(userAccount: string, dappAccount: string, origin: string): Promise<PoWCaptcha> {
```
forgetso, line 176: There should be some limitation on getting challenges. No challenge should be returned if the dapp does not have the POW product.

forgetso: Same goes for any of the other API calls relating to POW. Checks basically want to be

- does the app have POW?
- is the origin in the request the same as the registered dapp origin in the contract

Don't implement the origin check for now.

forgetso (changes requested):

demos/client-bundle-example/src/index.html
HughParry: Yes, for a follow up PR though to keep size reasonable. Added to this sprint

https://github.com/orgs/prosopo/projects/4/views/1?pane=issue&itemId=54753652

packages/procaptcha-bundle/src/index.tsx
HughParry: Fixed in latest push

demos/client-pow-example/env.development
HughParry: Changed to 9234 as lowest available port in the established pattern

demos/client-pow-example/vite.config.ts
HughParry: changed to 9234

packages/provider/src/tasks/tasks.ts
HughParry: To be resolved in 

https://github.com/orgs/prosopo/projects/4/views/1?pane=issue&itemId=54757049
https://github.com/orgs/prosopo/projects/4/views/1?pane=issue&itemId=54757136

That is unless you want the feature checking in this pr? Relates to https://github.com/orgs/prosopo/projects/4/views/1?pane=issue&itemId=54753652, which I think is a v high priority follow up pr imo

packages/database/src/databases/mongo.ts
```diff
@@ -444,6 +447,103 @@ export class ProsopoDatabase extends AsyncFactory implements Database {
         }
     }
 
+    /**
+     * @description Adds a new PoW Captcha record to the database.
+     * @param {string} challenge The challenge string for the captcha.
+     * @param {boolean} checked Indicates if the captcha has been checked.
+     * @returns {Promise<void>} A promise that resolves when the record is added.
+     */
+    async addPowCaptchaRecord(challenge: string, checked: boolean): Promise<void> {
```
forgetso: Please modify function name to `storePowCaptchaRecord` inline with the naming convention (`get`, `store`, `update`)

packages/procaptcha-pow/.npmignore
```diff
@@ -0,0 +1,17 @@
+/node_modules/
+/src/
+/tests/
+/artifacts/
```
forgetso: Is this file required?

packages/procaptcha/src/types/client.ts
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 import { CaptchaSolutionCommitmentId } from '@prosopo/api'
-import { CaptchaSolutionResponse } from '../types/api.js'
+import { CaptchaSolutionResponse } from '../../../types/src/procaptcha/api.js'
```
forgetso: This looks incorrect. Should it be `@prosopo/types` ?

packages/provider/src/tasks/tasks.ts
forgetso: yeah feature checking via the contract. Leave for now.

forgetso (changes requested):
Some small comments regarding naming

forgetso (approved):

PR #1070, opened 2024-02-28
Flux CLI

forgetso:


dev/flux/.npmignore
goastler, line 1: why does this need its own npmignore?

goastler (changes requested):

dev/flux/.npmignore
forgetso: I just copied it from another package, will delete.

goastler (approved):
:rocket:

PR #1078, opened 2024-03-04
Collector to send to mongodb for dev environments only

HughParry:


forgetso (dismissed):

packages/database/src/eventsDatabase/eventsDatabase.ts
```diff
@@ -36,47 +36,17 @@ try {
...
-    return new Promise((resolve, reject) => {
-        const connection = mongoose.createConnection(atlasUri)
-        for (let attempt = 1; attempt <= MAX_RETRIES; attempt++) {
-            connection.once('open', resolve).on('error', (e) => {
-                logger.warn(`Mongoose connection error`)
-                logger.error(e)
+    await mongoose
+        .connect(atlasUri)
+        .then(() => console.log('Connected to MongoDB Atlas'))
+        .catch((err) => console.error('Error connecting to MongoDB:', err))
```
goastler: this will catch the error and proceed onwards. Is this what you intended?

packages/procaptcha/src/modules/Manager.ts
```diff
@@ -597,12 +596,34 @@ export function Manager(
...
         }
-        const providerApi = await loadProviderApi(providerUrl)
-        await providerApi.submitUserEvents(events, getAccount().account.address)
+        console.log('Submitting events to provider', events)
+
+        let account = ''
+        try {
+            account = getAccount().account.address
+        } catch (e) {
+            console.error(e)
```
goastler, line 624: this will also proceed onwards with address as '', which means we'll have several db records with no account if something went wrong here. Not sure if that's a problem, just flagging it

packages/procaptcha/src/modules/collector.ts
```diff
@@ -1,6 +1,6 @@
 import { ProsopoKeyboardEvent, ProsopoMouseEvent, ProsopoTouchEvent } from '@prosopo/types'
 
-const COLLECTOR_LIMIT = 1000
+const COLLECTOR_LIMIT = 10000
```
goastler, line 3: we need to check what kind of db storage this takes. Is this the first 10k events which fire, then drop the rest?

packages/provider/src/tasks/tasks.ts
```diff
@@ -576,9 +576,10 @@ export class Tasks {
     }
 
     async saveCaptchaEvent(events: StoredEvents, accountId: string) {
-        if (!this.config.mongoAtlasUri) {
+        if (!this.config.devOnlyWatchEvents || !this.config.mongoEventsUri) {
+            console.log('Dev watch events not set to true, not saving events')
             return
         }
-        await saveCaptchaEvent(events, accountId, this.config.mongoAtlasUri)
+        await saveCaptchaEvent(events, accountId, this.config.mongoEventsUri)
```
goastler, line 718: are we changing the place we submit captchas too also?

goastler (dismissed):

packages/procaptcha/src/modules/collector.ts
HughParry: it's most recent 10k events - depends on polling rate of your mouse and how much you wiggle it

packages/provider/src/tasks/tasks.ts
HughParry: This is just terminology change for clarity

packages/procaptcha/src/modules/Manager.ts
HughParry: Yeah agreed - more interested in getting data into the db when getAccount() fails (which can happen in a dev environment)

packages/database/src/eventsDatabase/eventsDatabase.ts
HughParry: Yeah I don't want the provider to fall over just because it can't emit user events to mongo

packages/procaptcha/src/modules/Manager.ts
goastler: sure, makes sense :)

packages/database/src/eventsDatabase/eventsDatabase.ts
goastler: won't the next lines fallover because the db isn't connected?

packages/provider/src/tasks/tasks.ts
goastler: okay cool, no prob then :)

packages/procaptcha/src/modules/collector.ts
goastler: okay and how big are those when fully generated? just wondering how much storage this will take really

forgetso (approved):

forgetso (approved):

PR #1081, opened 2024-03-05
Make flux an executable and improve auth script

forgetso:
Enable running

```
npx flux auth --app ProcaptchaJavascriptServer
```

from the root of captcha


dev/flux/src/index.ts
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env node
```
goastler, line 1: I _think_ you can specify a binary file to run in the pkg json rather than adding this. Then the cmd would become `npx @prosopo/flux ...` https://docs.npmjs.com/cli/v10/configuring-npm/package-json#bin

no problem with doing both too tho!

goastler (dismissed):

dev/flux/src/index.ts
forgetso: You need both it seems :)

goastler (changes requested):
see slack

goastler (approved):

PR #1088, opened 2024-03-07
Rename test dirs

### LINKED ISSUE #1089: change test dirs to __tests__
this is jest convention and has lots of ubiquitous integration with other tools by default, e.g. typedoc

goastler:
`__tests__` is jest convention. We should use this as it integrates nicely with tools, and I need the tests to be in a well defined place to ignore them with typedoc

fixes prosopo/captcha-private#655


goastler:
This no longer renames the `tests` dir as no longer needed elsewhere. Instead, this moves the output of the test results to avoid git / watching tools from picking up changes to the src

+ introduces a cache dir for us to use

forgetso (approved):

PR #1093, opened 2024-03-13
Dynamic frictionless v1 impl

HughParry:


packages/procaptcha-react/README.md
```diff
@@ -19,7 +19,7 @@ See the [client example](https://github.com/prosopo/client-example) for a minima
 in a frontend app.
 
 ```jsx
-<Procaptcha config={config} callbacks={{ onAccountNotFound, onError, onHuman, onExpired }} />
+<Procaptcha config={config} callbacks={{ onError, onHuman, onExpired }} />
```
goastler, line 22: what happened to the callback here?

packages/types/src/procaptcha/manager.ts
```diff
@@ -41,3 +41,23 @@ export type ProcaptchaOutput = zInfer<typeof ProcaptchaOutputSchema>
...
+/**
+ * A set of callbacks called by Procaptcha on certain events. These are optional
+ * as the client can decide which events they wish to listen for.
+ */
+export type ProcaptchaCallbacks = Partial<ProcaptchaEvents>
+
+/**
+ * A list of all events which can occur during the Procaptcha process.
+ */
+export interface ProcaptchaEvents {
```
goastler, line 54: where were these before being moved to here?

goastler (changes requested):

packages/procaptcha-react/README.md
HughParry: It doesn't do anything different to the error callback, refactored so that it's now treated as a regular error

packages/types/src/procaptcha/manager.ts
HughParry: Defined in Procaptcha package. Made it a nightmare for code splitting and reusing ProcaptchaEvents outside of image captcha (which wasn't really much of an issue prior to pow/frictionless)

Basically just keeping types in the types pkg and avoiding circular deps :)

packages/procaptcha-react/README.md
goastler: It doesn't, sure, but the idea was to discern the case where an account can't be found from a general error. The former may need to be raised in a different way to the user/site admin. I know we just do the same thing in our examples, but I think client usage will likely be different

You can leave it like this if you like, as I guess it doesn't offer anything extra right now. Have you removed it from the bundles + other demos?

packages/types/src/procaptcha/manager.ts
goastler: okay sounds good

packages/procaptcha-react/README.md
HughParry: It's removed from everywhere yep.

I was thinking about this, and I'm not sure that an account not found error is any more special than the other error types really (e.g. extension not found, no accounts found, network not found, the various no/invalid captcha errors, sitekey errors, etc)

goastler: I concur good sir

goastler (approved):

PR #1095, opened 2024-03-13
Server verification test

forgetso:


demos/client-example-server/src/controllers/auth.ts
```diff
@@ -53,7 +53,7 @@ const signup = async (
         if (dbUser) {
             return res.status(409).json({ message: 'email already exists' })
         }
-
+        console.log('payload', payload)
```
goastler, line 56: intentional debug?

demos/cypress-shared/cypress/e2e/correct.captcha.cy.ts
```diff
@@ -30,7 +30,6 @@ describe('Captchas', () => {
                 context: { datasetWithSolutionHashes },
             })
         }
-        cy.intercept('/dummy').as('dummy')
```
goastler, line 33: why has this been removed?

packages/cli/src/cli.ts
```diff
@@ -32,6 +32,16 @@ async function main() {
         unsolved: { count: 0 },
     })
 
+    if (config.devOnlyWatchEvents) {
```
goastler, line 35: nice!

goastler (changes requested):

demos/client-example-server/src/controllers/auth.ts
forgetso: Yep, this is to help identify what's gone wrong when you're running locally. This is just a demo package.

demos/cypress-shared/cypress/e2e/correct.captcha.cy.ts
forgetso: I think it was just copied from stackoverflow by mistake. We don't have any API calls with `dummy` in them!

goastler (dismissed):

goastler (approved):

PR #1110, opened 2024-03-19
Procaptcha common

forgetso:
I've made a start on de-duplicating the procaptcha stuff and moved all procaptcha types into the types package.

- This allows us to de-duplicate a bunch of types.
- The state is now defined in procaptcha-common, which has the knock-on effect of requiring an interface for various classes in types

As it stands, we don't have any functionality that appends a payload to the result of a PoW captcha and this PR will allow me to develop that next (e.g. there is currently no server verification for Frictionless or Pow).


packages/procaptcha-common/src/events.ts
```diff
@@ -0,0 +1,54 @@
...
+
+const alertError = (error: ProsopoError) => {
+    alert(error.message)
+}
+
+export const getDefaultEvents = (
+    stateUpdater: ProcaptchaStateUpdateFn,
+    state: ProcaptchaState,
+    callbacks: ProcaptchaCallbacks
+): ProcaptchaEvents => {
```
HughParry: remove braces and return

packages/procaptcha-common/src/state.ts
```diff
@@ -0,0 +1,113 @@
...
+    ProcaptchaState,
+    ProcaptchaStateUpdateFn,
+    ProsopoCaptchaApiInterface,
+    TCaptchaSubmitResult,
+} from '@prosopo/types'
+
+type useRefType = <T>(defaultValue: T) => { current: T }
+type useStateType = <T>(defaultValue: T) => [T, (value: T) => void]
+
+export const buildUpdateState = (state: ProcaptchaState, onStateUpdate: ProcaptchaStateUpdateFn) => {
```
HughParry: This can just be

```
export const buildUpdateState = (state: ProcaptchaState, onStateUpdate: ProcaptchaStateUpdateFn) => 
    (nextState: Partial<ProcaptchaState>) => {
        Object.assign(state, nextState);
        onStateUpdate(nextState);
    };
```

Also quite noisy comments here

packages/procaptcha/src/modules/Manager.ts
```diff
@@ -267,35 +226,32 @@ export function Manager(
...
             providerApi = await loadProviderApi(providerUrl)
-            console.log('providerApi', providerApi)
+
             // get the captcha challenge and begin the challenge
             const captchaApi = await loadCaptchaApi(contract, getRandomProviderResponse, providerApi)
 
-            console.log('captchaApi', captchaApi)
-            const challenge: GetCaptchaResponse = await captchaApi.getCaptchaChallenge()
-            console.log('challenge', challenge)
+            const challenge: CaptchaResponseBody = await captchaApi.getCaptchaChallenge()
```
HughParry: Why explicitly typed? Should be inferred from captchaApi

HughParry (changes requested):

HughParry (approved):

PR #1116, opened 2024-03-20
fix typo in alice address

goastler:


goastler:
nvm, typo made elsewhere, this address was correct for alice

PR #1118, opened 2024-03-23
Arm64

forgetso:
Instructions and a compose file for getting working on Pi. Solutions script is removed for easier bundling.


HughParry (approved):

HughParry:
Note:

> This referrer extract thing is nice to have but would you mind moving it to the scripts package and adding as a script that calls the flux package to get the logs? I've promoted the flux package in the flux discord today and I don't want it filling up with custom prosopo code.

PR #1122, opened 2024-03-24
Add version to frontend vite bundling process

forgetso:


forgetso:
Closing in favour of adding a CICD step that populates version when we are releasing.

PR #1124, opened 2024-03-25
Tx queue

forgetso:
Contains:
- [x] A queue for handling transactions
- [x] A script for manually registering and funding dapps from the mongo db


dev/config/src/projectInfo.ts
```diff
@@ -164,6 +164,10 @@ export const getTypesPkgDir = () => {
     return `${getPackagesDir()}/types`
 }
 
+export const getTxPkgDir = () => {
```
HughParry: remove braces from func that immediately returns

dev/scripts/src/contract/fundDapps.ts
```diff
@@ -0,0 +1,124 @@
...
+    let abi: Abi
+    if (abiPath === undefined) {
+        log.info('Using abi from captcha contract')
+        abi = new Abi(ContractAbi)
+    } else {
+        log.info('ABI Path', abiPath)
+        abi = await AbiJSON(path.resolve(abiPath))
+    }
+
+    if (!atlasUri) {
```
HughParry: Can we move this to the top of the func so it fails fast

dev/scripts/src/contract/fundDapps.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    const dapps = await EmailsModel.find({}).exec()
+    log.info(dapps)
+
+    const config = defaultConfig()
+    const network = config.networks[config.defaultNetwork]
+    const secret = config.account.secret
+    const pair = await getPairAsync(network, secret)
+    const env = new ProviderEnvironment(config, pair)
+    await env.isReady()
```
HughParry, line 96: 91-96 is a pretty common pattern for us - should we impl a getDefaultEnv() that does this for us?

dev/scripts/src/setup/dapp.ts
```diff
@@ -12,14 +12,64 @@
...
+                        )
+                        .then((result) => {
+                            log.debug('Transaction added to queue', result)
+                        })
+                })
+            } catch (err) {
+                reject(err)
+            }
+        } else {
+            reject(new ProsopoContractError('CONTRACT.INVALID_METHOD'))
```
HughParry: Can we add the method to the context on this error please

packages/contract/src/contract/helpers.ts
```diff
@@ -74,20 +74,19 @@ export function encodeStringArgs(abi: Abi, methodObj: AbiMessage, args: any[]):
...
             const responseOk = out.asOk
             if (responseOk.isErr) {
-                throw new ProsopoContractError('CONTRACT.QUERY_ERROR', {
-                    context: { error: responseOk.toPrimitive().err.toString(), contractMethodName },
-                })
+                return responseOk.toPrimitive().err.toString()
             }
         }
     }
+    return undefined
```
HughParry: Potentially confusing when debugging? Could we throw here or return something other than undefined that explains that we've tried to get the contract error but there isn't one?

packages/contract/src/contract/interface.ts
```diff
@@ -146,6 +149,107 @@ export class ProsopoCaptchaContract extends Contract implements IProsopoCaptchaC
...
+
+        return this.abi.registry.createTypeUnsafe(
+            message.returnType.lookupName || message.returnType.type,
+            [outcome.result.asOk.data.toU8a(true)],
+            { isPedantic: true }
+        )
+    }
+
+    private argDecoder(arg: any): string {
+        let decoded = arg.toString()
```
HughParry: not a fan of the use of `let` here where not required

Personally would prefer:

`if (Array.isArray(arg)) {
return
...
return arg.toString()`

packages/contract/src/contract/interface.ts
```diff
@@ -205,6 +311,17 @@ export class ProsopoCaptchaContract extends Contract implements IProsopoCaptchaC
         }
     }
 
+    /** Get the contract method from the ABI
+     * @return the contract method object
+     */
+    getContractMethod(contractMethodName: string): AbiMessage {
+        const methodObj = this.contract?.abi.messages.filter((obj) => obj.method === contractMethodName)[0]
+        if (methodObj !== undefined) {
+            return methodObj as unknown as AbiMessage
```
HughParry, line 320: i hate polkadotjs

packages/provider/src/tests/accounts.ts
```diff
@@ -16,15 +16,16 @@ import { IDappAccount, IProviderAccount } from '@prosopo/types'
...
-    if (account[2]) {
-        return account[2]
+export const accountMnemonic = (account: TestAccount) => account.mnemonic
+export const accountAddress = (account: TestAccount) => account.address
+export const accountContract = function (account: TestAccount): string {
+    if (account.contractAddress) {
+        return account.contractAddress
     }
-    throw new Error(`Account ${account[1]} does not have a contract`)
+    throw new Error(`Account ${account} does not have a contract`)
```
HughParry: Are we happy to use Error instead of some sort of ProsopoError since this is in tests?

packages/provider/src/tests/dataUtils/funds.ts
```diff
@@ -66,39 +68,69 @@ export async function sendFunds(
...
+    if (!txQueue) {
+        // eslint-disable-next-line no-async-promise-executor
+        result = new Promise<ISubmittableResult>(async (resolve, reject) => {
+            const unsub = await api.tx.balances
+                .transferAllowDeath(address, amount)
+                .signAndSend(pair, { nonce }, (result: ISubmittableResult) => {
+                    if (result.status.isInBlock || result.status.isFinalized) {
+                        result.events
+                            .filter(({ event: { section } }: any): boolean => section === 'system')
+                            .forEach((event): void => {
```
HughParry: forEach never returns anything - shouldn't need to type it with `void`

packages/provider/src/tests/dataUtils/funds.ts
```diff
@@ -66,39 +68,69 @@ export async function sendFunds(
...
+            context: {
+                error: submittableResult.dispatchError
+                    ? submittableResult.dispatchError.toHuman()
+                    : submittableResult.toHuman(),
+            },
+        })
+    }
+    if (submittableResult.status.isFuture) {
+        const block = await api.rpc.chain.getBlock()
+        console.log('Current block', JSON.stringify(block))
```
HughParry: how come console.log + env.logger.debug? are they logging in different environments?

HughParry (changes requested):

dev/scripts/src/contract/fundDapps.ts
forgetso: Good spot. Definitely shouldn't be an optional param.

dev/scripts/src/contract/fundDapps.ts
forgetso: We used to pass in a mnemonic to the env but this isn't good from a security point of view. We should get rid of the secret/mnemonic as soon as possible, hence the pair stuff sitting outside. `getDefaultEnv()` sounds like a good idea. I'll add a follow up ticket for it.

packages/contract/src/contract/interface.ts
forgetso: Good spot.

packages/contract/src/contract/interface.ts
forgetso: We actually have a way to type this when the zod replacement is up and running :joy:

packages/provider/src/tests/accounts.ts
forgetso: it's just old code. I'll update.

packages/provider/src/tests/dataUtils/funds.ts
forgetso: Shouldn't be there. Removed.

HughParry (approved):

PR #1127, opened 2024-03-26
Tsconfig reference checker

goastler:
This checks whether the deps listed in pkg jsons match up with tsconfig jsons. We need this to check tsconfigs when using a this repo as a submodule


forgetso:
As mentioned, this will fail if the package.json is not up to date so we probably need a follow up issue to resolve these situations.

forgetso (approved):

PR #1134, opened 2024-04-04
Move test configs to individual packages

forgetso:
- Moves test configs to individual packages that import from one shared test config in dev/config.
- Adds dotenv dev dependency to each package using a vite.test.config.ts.
- Expects a `.env.{NODE_ENV}` file to be present at the root of the package or at the root of the repository and errors if it is missing
- Runs all tests from root with workspaces command `"test": "npm run --workspaces --if-present test",`


package.json
```diff
@@ -44,7 +44,7 @@
         "start:all": "npx concurrently \"npm run start:server\" \"npm run start:provider\" \"npm run start:demo\" \"npm run start:bundle\" --kill-others",
         "start:all:pow": "npx concurrently \"npm run start:server\" \"npm run start:provider\" \"npm run start:demo:pow\" \"npm run start:bundle\" --kill-others",
         "populate-data": "npm -w @prosopo/provider run populate-data",
-        "test": "npm -w @prosopo/scripts run test",
+        "test": "npm run --workspaces --if-present test",
```
goastler: rogue `--if-present`?

goastler (changes requested):

package.json
forgetso: Removed.

goastler (dismissed):

goastler (approved):
:rocket:

PR #1154, opened 2024-04-18
Vite Plugin - Watch Workspace

forgetso:
Adds a plugin that will watch local workspace dependencies for changes, rebuilding when they change.

- [x] watches all files in all packages in all workspaces
- [x] excludes current package where vite.config.ts is located (these are already watched)
- [x] has optional excludes via ignore globs
- [x] works for JSX

Peek 2024-04-20 21-04


HughParry:
@forgetso ready to approve after discussed changes to the switch case + defaults

HughParry (dismissed):

HughParry (approved):

PR #1158, opened 2024-04-22
Adding default max verified time to frontend

HughParry:


packages/procaptcha/src/modules/Manager.ts
```diff
@@ -84,12 +84,6 @@ export function Manager(
     onStateUpdate: ProcaptchaStateUpdateFn,
     callbacks: ProcaptchaCallbacks
 ) {
-    // events are emitted at various points during the captcha process. These each have default behaviours below which can be overridden by the frontend using callbacks.
-
-    const alertError = (error: ProsopoError) => {
```
goastler, line 89: was this not used?

goastler (changes requested):

packages/procaptcha/src/modules/Manager.ts
HughParry: nope

goastler (approved):

PR #1192, opened 2024-04-30
change callback and elementId to element

sunlight7777777:


demos/client-bundle-example/src/frictionless.html
```diff
@@ -6,24 +6,26 @@
...
-            window.onCaptchaVerified = function (output) {
+            function onCaptchaVerified(output) {
                 console.log('Captcha verified, output: ' + JSON.stringify(output))
             }
         </script>
         <script
             id="procaptchaScript"
             type="module"
-            src="./procaptcha.bundle.js?onload=captchaLoaded&render=explicit"
+            src="./assets/procaptcha.bundle.js?onload=captchaLoaded&render=explicit"
```
goastler, line 28: have you changed where the bundle lives? if so, we'll need to change other scripts to output to that dir

goastler (dismissed):

demos/client-bundle-example/src/jsBundleTest.html
```diff
@@ -15,7 +15,7 @@
             <div
                 class="procaptcha"
                 data-theme="light"
-                data-sitekey="5HUBceb4Du6dvMA9BiwN5VzUrzUsX9Zp7z7nSR2cC1TCv5jg"
+                data-sitekey="5HGjWAeFDfFCWPsjFQdVV2Msvz2XtMktvgocEZcCj68kUMaw"
```
HughParry: I don't think this needs to change

packages/procaptcha-bundle/src/index.tsx
```diff
@@ -195,12 +195,12 @@ const implicitRender = () => {
 }
 
 // Explicit render for targeting specific elements
-export const render = (elementId: string, renderOptions: ProcaptchaRenderOptions) => {
+export const render = (element: Element, renderOptions: ProcaptchaRenderOptions) => {
     const siteKey = renderOptions.siteKey
-    const element = document.getElementById(elementId)
+    // const element = document.getElementById(elementId)
```
HughParry: Please remove this instead of commenting out

packages/procaptcha-bundle/src/index.tsx
```diff
@@ -195,12 +195,12 @@ const implicitRender = () => {
...
 
     if (!element) {
-        console.error('Element not found:', elementId)
+        console.error('Element not found:', element)
```
HughParry: I'm not sure that this code path can be reached - the element is passed to render, which I believe guarantees that it exists

packages/procaptcha-bundle/src/index.tsx
```diff
@@ -242,4 +242,4 @@ if (onloadUrlCallback) {
     getProcaptchaScript()?.addEventListener('load', () => {
         ready(onloadCallback)
     })
-}
+}
```
HughParry: Looks like a linting issue here

packages/procaptcha-bundle/src/index.tsx
```diff
@@ -111,8 +111,8 @@ const renderLogic = (
     config: ProcaptchaClientConfigInput,
     renderOptions?: ProcaptchaRenderOptions
 ) => {
-    elements.forEach((element) => {
-        const callbackName = renderOptions?.callback || element.getAttribute('data-callback')
+    elements.forEach((element) => { // || element.getAttribute('data-callback')
```
HughParry: Remove this instead of commenting out please

package.json
```diff
@@ -115,7 +115,7 @@
         "typedoc-plugin-mdn-links": "^3.1.16",
         "typedoc-plugin-missing-exports": "^2.2.0",
         "typedoc-plugin-zod": "^1.1.2",
-        "typescript": "5.1.6"
+        "typescript": "^5.1.6"
```
HughParry: Why?

HughParry (dismissed):

demos/client-bundle-example/src/jsBundleTest.html
sunlight7777777: Yes, this is my mistake
Sorry

packages/procaptcha-bundle/src/index.tsx
sunlight7777777: Okay, I got it

packages/procaptcha-bundle/src/index.tsx
sunlight7777777: Okay, I will remove

packages/procaptcha-bundle/src/index.tsx
sunlight7777777: Okay, I got it

demos/client-bundle-example/src/jsBundleTest.html
sunlight7777777: I checked this again but I didn't change this

HughParry: Will have changed as part of a build step then - fine for it to change

packages/procaptcha-bundle/src/index.tsx
sunlight7777777: I will fix it

package.json
sunlight7777777: It was not installed yesterday

demos/client-bundle-example/src/frictionless.html
sunlight7777777: Should I change this to origin again?

forgetso: The `src` was previously incorrect. This change fixes the page.

> Should I change this to origin again?

No, leave this as assets

package.json
HughParry: It was installed - the ^ sign means that it will allow for minor version bumps (e.g. it would be happy to install 5.1.7, even though you've specified 5.1.6). Not having it means the version is pinned to 5.1.6.

Lets keep ts pinned to 5.1.6 for now so that type behaviour is stable

demos/client-bundle-example/src/frictionless.html
sunlight7777777: I changed this and pushed again now...

demos/client-bundle-example/src/jsBundleTest.html
forgetso: Sorry guys, this actually needs to remain as `5HUBceb4Du6dvMA9BiwN5VzUrzUsX9Zp7z7nSR2cC1TCv5jg`. The reason is the tests rely on this *live* site key 

https://github.com/prosopo/captcha/blob/a4c47396fa38d01661695d3a55d8b0a06db776c9/.github/workflows/publish.yml#L423

`5HUBceb4Du6dvMA9BiwN5VzUrzUsX9Zp7z7nSR2cC1TCv5jg` is registered as a dapp/sitekey in our live rococo contract.

Unfortunately, the dev setup script (`npm run setup`) searches for these site key values and replaces them with the development site key. We could fix this by doing the search replace during the CICD [here](https://github.com/prosopo/captcha/blob/a4c47396fa38d01661695d3a55d8b0a06db776c9/.github/workflows/publish.yml#L423) and replacing the value with a github secret.

forgetso: For now, @Vengi717, please revert this change.

packages/procaptcha-bundle/package.json
```diff
@@ -62,7 +62,7 @@
         "@rollup/plugin-typescript": "^11.1.2",
         "@vitejs/plugin-react": "^4.2.1",
         "tslib": "2.6.2",
-        "typescript": "5.1.6",
+        "typescript": "^5.1.6",
```
forgetso, line 65: Please pin

forgetso (changes requested):
Couple of extra comments from me.

forgetso (changes requested):
Comments added

forgetso (approved):
Great work.

PR #1202, opened 2024-05-07
bump versions to 0.3.40

forgetso:


HughParry:
closing until bumping with more stuff in, will run from GHA again

PR #1203, opened 2024-05-08
Allowing setting callbacks via string or explicit function

HughParry:


demos/client-bundle-example/src/jsBundleTest.html
```diff
@@ -15,7 +15,7 @@
             <div
                 class="procaptcha"
                 data-theme="light"
-                data-sitekey="5HUBceb4Du6dvMA9BiwN5VzUrzUsX9Zp7z7nSR2cC1TCv5jg"
+                data-sitekey="5HGjWAeFDfFCWPsjFQdVV2Msvz2XtMktvgocEZcCj68kUMaw"
```
forgetso: Please don't revert this or add CICD `sed` step in to set it to a secret github var with this value `5HUBceb4Du6dvMA9BiwN5VzUrzUsX9Zp7z7nSR2cC1TCv5jg` at the point where its required.

forgetso (dismissed):

demos/client-bundle-example/src/jsBundleTest.html
forgetso: *change

PR #1234, opened 2024-05-20
Prosopo api error error code fixes

HughParry:


packages/provider/src/api/authMiddleware.ts
```diff
@@ -46,12 +46,14 @@ const extractHeaders = (req: Request) => {
 
     if (!signature || !blocknumber) {
         throw new ProsopoApiError('CONTRACT.INVALID_DATA_FORMAT', {
-            context: { error: 'Missing signature or block number' },
+            context: { error: 'Missing signature or block number', code: 400 },
```
goastler, line 49: you've used status on other lines, is it status or code?

goastler (changes requested):

packages/provider/src/api/authMiddleware.ts
HughParry: Should be code everywhere. Status incorrectly used in Flux error handler. Good catch

goastler (approved):

goastler:
@prosoponator u alive?

goastler:
@prosoponator approve

prosoponator (approved):

goastler:
@prosoponator disapprove

PR #1239, opened 2024-05-21
lock provider_configure() to admin only access

goastler:
prevents a rococo whale messing with us. Locks the contract so only admins can mutate/register providers. This is a stop gap solution until we're on a live chain / have proper staking setup


goastler:
closing due to dropping contract

PR #1250, opened 2024-05-27
Add Astar as a network option

forgetso:
Creates a version of the code that runs on [prosopo/substrate-contracts-node:v0.41.18dp](https://hub.docker.com/layers/prosopo/substrate-contracts-node/v0.41.18dp/images/sha256-d2dbd597e6feff8b9b26a45ace788adf4e38a4ccf4bfee34ab6fa598c228eeca?context=repo), which to mimic Astar's 18 decimal point environment.


dev/config/src/provider/test.ts
```diff
@@ -40,7 +40,7 @@ export default function getTestConfig() {
         },
         networks: {
             development: {
-                endpoint: 'ws://localhost:9944',
+                endpoint: ['ws://localhost:9944'],
```
goastler, line 43: why an array now?

packages/provider/src/tests/tasks/tasks.test.ts
```diff
@@ -97,7 +96,13 @@ describe.sequential('CONTRACT TASKS', async function (): Promise<void> {
             context.env.getContractInterface().query.getProviderStakeThreshold,
             context.env.getContractInterface().query
         )()
-        context.providerStakeThreshold = new BN((await promiseStakeDefault).toNumber())
+
+        const dappStakeDefault: Promise<ReturnNumber> = wrapQuery(
+            context.env.getContractInterface().query.getDappStakeThreshold,
+            context.env.getContractInterface().query
+        )()
+        context.providerStakeThreshold = (await promiseStakeDefault).rawNumber
```
goastler, line 104: what does rawNumber do here?

protocol/contracts/common/src/lib.rs
```diff
@@ -73,8 +73,8 @@ pub mod common {
         /// Get the git commit id from when this contract was built
         pub fn get_git_commit_id() -> [u8; 20] {
             let env_git_commit_id: [u8; 20] = [
-                121, 218, 22, 102, 102, 85, 89, 222, 84, 212, 25, 77, 140, 7, 242, 242, 224, 116,
-                71, 45,
+                167, 193, 167, 170, 152, 101, 3, 78, 120, 176, 222, 146, 186, 108, 111, 84, 189,
```
goastler, line 76: what account is this?

goastler (changes requested):

dev/config/src/provider/test.ts
forgetso: Because PJS allows us to specify multiple endpoints (presumably for failover) so I've added as many endpoints as possible for Astar.

packages/provider/src/tests/tasks/tasks.test.ts
forgetso: Returns a `BN` which can handle the large numbers on Astar

protocol/contracts/common/src/lib.rs
forgetso: This is the git commit. I couldn't get the docker building to work via node so not sure if this will be correct. Presumably, it would have failed the typechain step if it was wrong. Is that right?

HughParry (approved):

protocol/contracts/common/src/lib.rs
goastler: whoops, sorry thought that was an account rather than the git commit lol

goastler (approved):

PR #1251, opened 2024-05-28
Move to Astar

HughParry:


dev/config/package.json
```diff
@@ -43,6 +43,7 @@
         "@babel/plugin-transform-react-jsx": "^7.24.6",
         "@babel/plugin-transform-runtime": "^7.24.3",
         "@babel/preset-env": "^7.24.5",
+        "@polkadot/dev": "^0.76.11",
```
forgetso: Is this required?

dev/scripts/src/contract/transferContract.ts
```diff
@@ -0,0 +1,88 @@
...
+    const config = defaultConfig()
+    const network = config.networks[config.defaultNetwork]
+    const secret = config.account.secret
+    const pair = await getPairAsync(network, secret)
+    const env = new ProviderEnvironment(config, pair)
+    await env.isReady()
+    const tasks = new Tasks(env)
+
+    addresses.forEach(async (address) => {
+        await tasks.contract.query.dappRegister(address, DappPayee.dapp)
```
forgetso: Please add a value of the `dappStakeThreshold` here so that the dapp is activated on registration. Otherwise, we need to do this as a follow up when the user clicks the verify link (maybe that's better?).

forgetso (dismissed):
Couple of small questions.

dev/config/package.json
HughParry: nah

HughParry:
@prosoponator approve

prosoponator (approved):

PR #1260, opened 2024-06-01
Fix env files and add signing CLI command to Flux package

forgetso:
Now that we're on Astar our env files need to be updated.

A signing mechanism has been created to make updating specifications on Flux easier.


forgetso:
@prosoponator approve

prosoponator (approved):

PR #1269, opened 2024-06-12
Only run lint and prettier on changed files locally

forgetso:
- [x] change local lint command
- [x]  remove old docs


package.json
```diff
@@ -52,21 +52,21 @@
         "eslint:workspace": "npm run eslint:cmd -- '.*.*' '*.*' .vscode .github docker",
         "eslint:cmd": "npx eslint --ignore-path .eslintignore --no-error-on-unmatched-pattern --cache --cache-location node_modules/.cache/eslint/.eslintcache",
         "eslint:contracts": "npm -w @prosopo/captcha-contract -w @prosopo/common-contract -w @prosopo/proxy-contract run eslint",
-        "eslint:fix": "npm run ws -- eslint:fix && npm run eslint:fix:workspace",
+        "eslint:fix": "npx eslint --ignore-path .eslintignore --no-error-on-unmatched-pattern --cache --cache-location node_modules/.cache/eslint/.eslintcache",
```
goastler, line 55: why has this changed?

package.json
```diff
@@ -52,21 +52,21 @@
...
         "eslint:fix:workspace": "npm run eslint:workspace -- --fix",
         "eslint:fix:contracts": "npm -w @prosopo/captcha-contract -w @prosopo/common-contract -w @prosopo/proxy-contract run eslint:fix",
         "prettier": "npm run ws -- prettier && npm run prettier:workspace",
         "prettier:workspace": "npm run prettier:cmd -- '.*.*' '*.*' .vscode .github docker",
         "prettier:cmd": "npx prettier --no-error-on-unmatched-pattern --check --ignore-path .eslintignore --ignore-unknown --cache --cache-location node_modules/.cache/prettier/.prettiercache",
         "prettier:contracts": "npm -w @prosopo/captcha-contract -w @prosopo/common-contract -w @prosopo/proxy-contract run prettier",
-        "prettier:fix": "npm run ws -- prettier:fix && npm run prettier:fix:workspace",
+        "prettier:fix": "npx prettier --no-error-on-unmatched-pattern --write --ignore-path .eslintignore --ignore-unknown --cache --cache-location node_modules/.cache/prettier/.prettiercache",
```
goastler, line 62: why has this changed?

goastler (changes requested):

package.json
forgetso: Because the linting is done as one command over the entire repository. It doesn't run in individual packages. We'd need to add extra ignore for when obf-bot is included.

package.json
forgetso: Same as above, the prettier command just runs once over a list of files instead of per package.

goastler (approved):

PR #1286, opened 2024-06-19
Flux updates

forgetso:
- [x] Make the CLI easier to use by not asking the user to specify `--app`. They can simply write the app name.
- [x] Add a CLI command for signing a payload
- [x] Add a CLI command for formatting the env file in the correct format for deploying on Flux
- [x] Add the first steps for updating env on enterprise nodes as a command


forgetso:
There will need to be a follow up PR in captcha-private once this is released.

goastler (approved):

PR #1289, opened 2024-06-25
Provider rate limiting

HughParry:


packages/cli/src/start.ts
```diff
@@ -60,6 +69,8 @@ export async function start(env?: ProviderEnvironment, admin?: boolean) {
     }
     await env.isReady()
 
+    console.log(env.config.rateLimits)
```
goastler: old console log?

packages/cli/src/start.ts
```diff
@@ -37,6 +39,13 @@ function startApi(env: ProviderEnvironment, admin = false): Server {
         apiApp.use(prosopoAdminRouter(env))
     }
 
+    // Rate limiting
+    const rateLimits = env.config.rateLimits
+    for (const [path, limit] of Object.entries(rateLimits)) {
+        const enumPath = path as CombinedApiPaths
+        apiApp.use(enumPath, rateLimit(limit))
+    }
+
```
goastler, line 48: should we set a default rate limit to all routes? I.e. '*' would be the route, and we'd set it to a generous rate limit which would allow reasonable usage but stop ddos. If we did this first, then set the rate limits per route then specific routes could have more stricter rate limits. Does that make sense?

goastler (changes requested):

packages/cli/src/start.ts
HughParry: Default rate limits are already defined though

```
export const ProviderDefaultRateLimits = {
    [ApiPaths.GetImageCaptchaChallenge]: { windowMs: 60000, limit: 30 },
    [ApiPaths.GetPowCaptchaChallenge]: { windowMs: 60000, limit: 60 },
    [ApiPaths.SubmitImageCaptchaSolution]: { windowMs: 60000, limit: 60 },
    [ApiPaths.SubmitPowCaptchaSolution]: { windowMs: 60000, limit: 60 },
    [ApiPaths.VerifyPowCaptchaSolution]: { windowMs: 60000, limit: 60 },
    [ApiPaths.VerifyImageCaptchaSolutionDapp]: { windowMs: 60000, limit: 60 },
    [ApiPaths.VerifyImageCaptchaSolutionUser]: { windowMs: 60000, limit: 60 },
    [ApiPaths.GetProviderStatus]: { windowMs: 60000, limit: 60 },
    [ApiPaths.GetProviderDetails]: { windowMs: 60000, limit: 60 },
    [ApiPaths.SubmitUserEvents]: { windowMs: 60000, limit: 60 },
    [AdminApiPaths.BatchCommit]: { windowMs: 60000, limit: 5 },
    [AdminApiPaths.UpdateDataset]: { windowMs: 60000, limit: 5 },
    [AdminApiPaths.ProviderDeregister]: { windowMs: 60000, limit: 1 },
    [AdminApiPaths.ProviderUpdate]: { windowMs: 60000, limit: 5 },
}
```

goastler (approved):

PR #1300, opened 2024-07-09
Remove contract calls from procaptcha

forgetso:


packages/contract/src/contract/block.ts
```diff
@@ -36,13 +36,15 @@ export const getCurrentBlockNumber = async (api: ApiPromise): Promise<number> =>
  * @param maxVerifiedTime
  * @param blockNumber
  */
-export const verifyRecency = async (api: ApiPromise, blockNumber: number, maxVerifiedTime: number) => {
+export const verifyRecency = async (challenge: string, maxVerifiedTime: number) => {
     // Get the current block number
```
forgetso, line 40: Change confusing comment

packages/procaptcha-pow/src/Services/Manager.ts
```diff
@@ -237,6 +209,30 @@ export const Manager = (
         }
     }
 
+    const getRandomActiveProvider = (): RandomProvider => {
+        const randomIntBetween = (min: number, max: number) => Math.floor(Math.random() * (max - min + 1) + min)
+
+        // TODO maybe add some signing of timestamp here by the current account and then pass the timestamp to the Provider
+        //  to ensure that the random selection was completed within a certain timeframe
+
+        const PROVIDERS = loadBalancer('development')
```
forgetso, line 217: Hard coded `development`?

packages/procaptcha/src/modules/Manager.ts
```diff
@@ -74,6 +68,30 @@ const getNetwork = (config: ProcaptchaClientConfigOutput) => {
     return network
 }
 
+const getRandomActiveProvider = (): RandomProvider => {
```
forgetso, line 68: Please make a shared function for this in procaptcha-common or something like that

packages/procaptcha/src/modules/Manager.ts
```diff
@@ -149,104 +169,26 @@ export function Manager(
...
-                    })
-                )
-                setValidChallengeTimeout()
-                return
-            }
-
-            // Check if there is a provider in local storage or get a random one from the contract
-            const procaptchaStorage = storage.getProcaptchaStorage()
-            let providerApi: ProviderApi
-            if (procaptchaStorage.providerUrl && procaptchaStorage.blockNumber) {
```
forgetso, line 186: You've removed cached image provider stuff. Is this intentional?

packages/procaptcha/src/modules/providers.ts
```diff
@@ -0,0 +1,59 @@
...
+import { EnvironmentTypes } from '@prosopo/types'
+
+type HardcodedProvider = {
+    address: string
+    url: string
+    datasetId: string
+    datasetIdContent: string
+}
+
+export const loadBalancer = (environment: EnvironmentTypes): HardcodedProvider[] => {
```
forgetso, line 24: Shared function please.

packages/provider/src/tasks/tasks.ts
```diff
@@ -264,7 +264,7 @@ export class Tasks {
             return false
         }
 
-        const [blocknumber, userAccount, challengeDappAccount] = challengeRecord.challenge.split(POW_SEPARATOR)
+        const [blocknumber, _, challengeDappAccount] = challengeRecord.challenge.split(POW_SEPARATOR)
```
forgetso, line 267: Change reference from `blocknumber` to `timestamp`

packages/provider/src/tasks/tasks.ts
```diff
@@ -359,14 +381,15 @@ export class Tasks {
                 userAccount: userAccount,
                 dappContract: dappAccount,
                 providerAccount: this.contract.pair.address,
-                datasetId: provider.datasetId.toString(),
+                datasetId,
                 status: CaptchaStatus.pending,
                 userSignature: Array.from(userSignature),
                 requestedAt: pendingRecord.requestedAtBlock, // TODO is this correct or should it be block number?
```
forgetso, line 387: We can probably drop this `requestedAtBlock` part.

packages/provider/src/tasks/tasks.ts
```diff
@@ -314,14 +315,10 @@ export class Tasks {
         dappAccount: string,
         requestHash: string,
         captchas: CaptchaSolution[],
-        signature: string // the signature to indicate ownership of account
+        signature: string, // the signature to indicate ownership of account
+        timestamp: string,
+        signedTimestamp: string
     ): Promise<DappUserSolutionResult> {
-        if (!(await this.dappIsActive(dappAccount))) {
```
forgetso, line 319: We will need to replace this with a call to the mongo client database checking that the app is active on one of our tiers. Please add as a follow up ticket.

packages/server/src/server.ts
```diff
@@ -208,14 +180,7 @@ export class ProsopoServer {
             // We don't care about recency with PoW challenges as they are single use, so just return the verified result
             return result.verified
         }
-        const recent = await verifyRecency((await this.getContractApi()).api, blockNumber, timeouts.image.cachedTimeout)
-
-        if (!recent) {
-            // bail early if the block is too old. This saves us calling the Provider.
```
forgetso, line 214: It's still worthwhile putting the equivalent check in here for signed timestamp, to avoid unnecessary calls to the provider.

packages/server/src/server.ts
```diff
@@ -232,26 +197,13 @@ export class ProsopoServer {
...
-            // If we have a providerURL and a blockNumber, we verify with the provider
+        const { providerUrl, blockNumber, challenge } = ProcaptchaOutputSchema.parse(payload)
 
+        if (providerUrl) {
             return await this.verifyProvider(token, blockNumber, this.config.timeouts, providerUrl, challenge)
         } else {
-            // If we don't have a providerURL, we verify with the contract
-            return await this.verifyContract(user, this.config.timeouts.contract.maxVerifiedTime)
+            // If we don't have a providerURL, something has gone deeply wrong
+            throw new ProsopoApiError('API.BAD_REQUEST', { context: { message: 'No provider URL' } })
```
forgetso, line 206: Don't error here, just return `false` and error log it.

forgetso (commented):
Looking really good. A few changes on first review and I'm going to start running it locally before re-reviewing.

PR #1301, opened 2024-07-19
Conditional build and test script

goastler:
Given a list of files that has changed, find which packages the files belong to and then find packages that depend on that package.


goastler:
Need to overhaul the workflows first then put this in. Depending on build times, this may no longer be required

goastler:
chose not to go ahead with conditional builds due to the risk of introducing bugs from builds being skipped / tests not being run due to indirect dependencies between pkgs

PR #1304, opened 2024-07-23
Fix cypress install step in cache workflow

goastler:
- [ ] add cypress as dev dep
- [ ] remove cypress install from cache workflow


goastler:
wrong branch

PR #1307, opened 2024-07-23
New linting

goastler:
- [x] quicker linting
- [x] use biome
- [x] lint better, e.g. use dedicated linting tools for yaml rather than a dodgy eslint plugin that doesn't lint fully
- [x] reconfigure linting commands in npm scripts
- [x] ~package specific linting, e.g. nested biome configs~ can be done, but gets messy. Leave it until we actually need this functionality for a specific pkg


goastler:
blocked by #1308

goastler:
wrong branch

PR #1320, opened 2024-08-09
polyfill node imports

goastler:


goastler:
had this working locally but not on gha for some reason...

goastler:
closing due to fix in another pr

PR #1336, opened 2024-08-22
Store something on each request. Update PoW types. Additional checks

forgetso:
We need to be able to see how many requests for Proof of Work captcha have been made

- [x] This PR changes the PoW generation mechanism to generate a record at the point of challenge creation. Various attributes are modified by subsequent requests to mark the stage the challenge is at (requested, userChecked, serverChecked).
- [x] The PoW types also receive a fairly big refactor.
- [x] The signatures are changed to be requestHash signatures for image captcha. Signature for PoW captcha remain as challenge (provider) and timestamp (user).
- [x] Types addded for get image challenge URL so that tests don't build if the URL changes
- [x] PoW and Image captcha records now inherit from a common type containing common statuses and a result object
- [x] Changes TranslationKeys to zod type for compatibility with zod schemas
- [x] add lastUpdatedTimestamp to captcha records so that scheduled tasks don't send the same data multiple times
- [x] store audit details of scheduled captcha sending task


packages/provider/src/api/captcha.ts
```diff
@@ -35,6 +35,8 @@ import express, { type Router } from "express";
 import { Tasks } from "../tasks/tasks.js";
 import { handleErrors } from "./errorHandler.js";
 
+const NO_IP_ADDRESS = "NO_IP_ADDRESS";
```
goastler: probably want to chuck 'as const' on the end of this for better typing

goastler (dismissed):

dev/scripts/src/setup/setup.ts
```diff
@@ -140,13 +139,13 @@ export async function setup(force: boolean) {
 
     env.logger.info(`Registering provider... ${defaultProvider.address}`);
 
-    defaultProvider.pair = await getPairAsync(network, providerSecret);
+    defaultProvider.pair = await getPairAsync(providerSecret);
 
     // If no PROSOPO_SITE_KEY is present, we will register a test account like //Eve.
```
HughParry: Stale comments are pretty confusing here

packages/api/src/api/ProviderApi.ts
```diff
@@ -31,73 +30,69 @@ import {
...
-          [ApiParams.user]: {
-            [ApiParams.requestHash]: userRequestHashSignature,
-          },
-          [ApiParams.provider]: {
-            [ApiParams.timestamp]: providerTimestampSignature,
-          },
+    const body: CaptchaSolutionBodyType = {
+      [ApiParams.user]: userAccount,
+      [ApiParams.dapp]: this.account,
+      captchas,
```
HughParry: all of these should be api params: blah

packages/provider/src/api/captcha.ts
```diff
@@ -49,47 +52,45 @@ export function prosopoRouter(env: ProviderEnvironment): Router {
...
+        await tasks.imgCaptchaManager.getRandomCaptchasAndRequestHash(
+          datasetId,
+          user,
+          req.ip || NO_IP_ADDRESS,
         );
-      }
-    },
-  );
+      const captchaResponse: CaptchaResponseBody = {
+        captchas: taskData.captchas.map((captcha: Captcha) => ({
```
HughParry: [ApiParams.captchas] or something

packages/provider/src/tasks/powCaptcha/powTasks.ts
```diff
@@ -75,57 +74,101 @@ export class PowCaptchaManager {
...
+    const challengeRecord =
+      await this.db.getPowCaptchaRecordByChallenge(challenge);
+
+    if (!challengeRecord) {
+      logger.debug("No record of this challenge");
+      // no record of this challenge
+      return false;
+    }
+
+    if (challengeRecord.ipAddress !== ipAddress) {
```
HughParry: If we could drop this for now please

packages/provider/src/tasks/powCaptcha/powTasksUtils.ts
```diff
@@ -31,19 +31,8 @@ export const checkPowSolution = (
...
-        ERROR: "Captcha solution is invalid",
-        failedFuncName: checkPowSolution.name,
-        nonce,
-        challenge,
-        difficulty,
-      },
-    });
-  }
+): boolean => {
+  return validateSolution(nonce, challenge, difficulty);
```
HughParry: can we drop this func now that it just passes to a different one?

packages/provider/src/tasks/powCaptcha/powTasksUtils.ts
```diff
@@ -72,15 +61,6 @@ export const checkPowSignature = (
...
-    throw new ProsopoContractError("CAPTCHA.INVALID_TIMESTAMP", {
-      context: {
-        ERROR: `Timestamp in which the Provider was selected must be within the last ${timeout / 1000} seconds`,
-        failedFuncName: checkRecentPowSolution.name,
-        challenge,
-      },
-    });
-  }
+): boolean => {
+  return verifyRecency(challenge, timeout);
```
HughParry: can we drop this func now that it just passes to a different one?

HughParry (commented):
few comments

dev/scripts/src/setup/setup.ts
forgetso: Removed

packages/provider/src/tasks/powCaptcha/powTasks.ts
forgetso: Dropped

HughParry (approved):

PR #1342, opened 2024-09-02
Add types to tests to filter by test type during testing

goastler:
e.g. `TEST_TYPE=unit npm run test`

can specify multiple, e.g. `TEST_TYPE=unit,integration npm run test`

if the `TEST_TYPE` env var is not set then all tests are included.


forgetso (approved):
🙌

PR #1343, opened 2024-09-02
security policy

goastler:


goastler:
@forgetso does this suffice for the time being?

forgetso (approved):

PR #1345, opened 2024-09-02
remove old demo/example pkgs

goastler:


HughParry:
Please impl the client frictionless example into the regular demo if you're removing it here

goastler:
sort npm scripts associated with these too

goastler:
- pow added
- npm scripts adjusted

PR #1348, opened 2024-09-04
Provider retry

HughParry:


package.json
```diff
@@ -37,6 +37,7 @@
 		"setup": "npm -w @prosopo/scripts run setup",
 		"setup:all": "npm run setup",
 		"start:provider": "npm -w @prosopo/cli run start",
+		"start:provider:dev": "npm -w @prosopo/cli run start:dev",
```
goastler: why not use NODE_ENV=development like other script do?

packages/cli/src/argv.ts
```diff
@@ -38,6 +39,7 @@ export function processArgs(
 	return yargs(hideBin(args))
 		.usage("Usage: $0 [global options] <command> [options]")
 		.option("api", { demand: false, default: false, type: "boolean" } as const)
+		.option("dev", { demand: false, default: false, type: "boolean" } as const)
```
goastler: use NODE_ENV

packages/cli/src/cli.ts
```diff
@@ -53,11 +53,20 @@ async function main() {
...
-			});
+		if (processedArgs.dev) {
+			await new ReloadingAPI(envPath, config, pair, processedArgs)
+				.startDev()
+				.then(() => {
+					log.info("Reloading API started...");
+				});
+		}
+		else {
+			await new ReloadingAPI(envPath, config, pair, processedArgs)
```
goastler, line 63: I can't see the different between the branches of the if statement here

packages/load-balancer/src/index.ts
```diff
@@ -18,77 +18,55 @@ export type HardcodedProvider = {
...
+	return Object.values(provider);
+};
+
+export const loadBalancer = async (
+	environment: EnvironmentTypes,
+): Promise<HardcodedProvider[]> => {
 	if (environment === "production") {
-		return [
+		const providers: hostedProviders = await fetch(
+			"https://provider-list.prosopo.io/",
```
goastler, line 36: question on slack

packages/procaptcha-common/src/providers.ts
```diff
@@ -0,0 +1,51 @@
...
+} from "@prosopo/types";
+import { at } from "@prosopo/util";
+
+export const getRandomActiveProvider = async (
+	config: ProcaptchaClientConfigOutput,
+): Promise<RandomProvider> => {
+	const randomIntBetween = (min: number, max: number) =>
+		Math.floor(Math.random() * (max - min + 1) + min);
+
+	// TODO maybe add some signing of timestamp here by the current account and then pass the timestamp to the Provider
```
goastler, line 27: I don't think this matters tbh. We're just spreading frontend load to the backend, we don't need to verify if it's been done properly bc the majority of traffic will have

packages/procaptcha-common/src/providers.ts
```diff
@@ -0,0 +1,51 @@
...
+	attemptCount: number,
+	retryMax: number,
+) => {
+	try {
+		await currentFn();
+	} catch (err) {
+		if (attemptCount >= retryMax) {
+			return stateReset();
+		}
+		console.error(err);
```
goastler, line 62: should be logged properly or thrown or something

packages/procaptcha-pow/src/services/Manager.ts
```diff
@@ -126,140 +128,129 @@ export const Manager = (
...
-			throw new ProsopoEnvError("GENERAL.ACCOUNT_NOT_FOUND", {
-				context: { error: "Account address has not been set for web3 mode" },
-			});
-		}
-
-		// get a random provider
-		const getRandomProviderResponse = getRandomActiveProvider();
-
-		const events = getDefaultEvents(onStateUpdate, state, callbacks);
+		await providerRetry(
```
goastler, line 131: I don't think this needs to be a fn which wraps everything. Imo it should do the provider retry at provider selection, i.e on the getRandomActiveProvider call. It can stay as is for now, but I think this should be simplified at some point bc it's rather confusing atm

goastler (changes requested):

packages/cli/src/cli.ts
HughParry: One calls startDev() the other calls start()

packages/load-balancer/src/index.ts
HughParry: resolved

packages/procaptcha-common/src/providers.ts
HughParry: This is just copying over a comment from @forgetso so leaving in for now

packages/cli/src/argv.ts
HughParry: done

package.json
HughParry: done

packages/procaptcha-common/src/providers.ts
HughParry: Adding error logs

packages/cli/src/cli.ts
goastler: ah , I'm blind

goastler (dismissed):

packages/procaptcha-pow/src/services/Manager.ts
HughParry: Resolved on slack

goastler (approved):

PR #1351, opened 2024-09-05
Portal settings

forgetso:
- [x] overhaul db package to use base mongo class that all dbs inherit
  - [x] remove the async constructor
- [x] add types for user settings in types package
- [x] add a domain validator to util 
- [x] 🐛 stop the provider process from connecting to the db lots of times
- [x] add a scheduled job to retrieve the client list from the client db
- [x] remove the user events API
- [x] put client related tasks in a ClientTaskManager
- [x] remove the batch commit stuff from the config

Follow up
- [ ] Create an API that triggers providers to look for new clients
- [ ] Use user settings in bot detection process (`domains`, `threshold`)
- [ ] Remove clients from provider database if no longer activated or present in the central list


dev/scripts/src/migrates/provider.db.2.0.1.to.2.0.2.ts
```diff
@@ -1,3 +1,11 @@
+import { loadEnv } from "@prosopo/dotenv";
```
HughParry, line 1: Dodgy linting change

packages/cli/src/cli.ts
```diff
@@ -65,6 +65,7 @@ async function main() {
 
 //if main process
 if (isMain(import.meta.url, "provider")) {
+	console.log("We're in the main process...");
```
HughParry: Need better debug here - relates to the dev vs non dev cli changes in another PR

goastler: drop

packages/cli/src/start.ts
```diff
@@ -63,28 +64,34 @@ function startApi(env: ProviderEnvironment, admin = false): Server {
 
 export async function start(env?: ProviderEnvironment, admin?: boolean) {
 	if (!env) {
+		console.log("Env undefined");
```
goastler: drop?

packages/database/src/databases/provider.ts
```diff
@@ -63,53 +65,99 @@ import {
...
+];
+
+export class ProviderDatabase extends MongoDatabase implements IDatabase {
+	tables = {} as Tables<TableNames>;
+
+	constructor(
 		url: string,
-		dbname: string,
-		logger: Logger,
+		dbname?: string,
```
goastler, line 137: why is this optional? surely you need a dbname?

packages/provider/src/schedulers/getClientList.ts
```diff
@@ -0,0 +1,56 @@
...
+// limitations under the License.
+
+import type { KeyringPair } from "@polkadot/keyring/types";
+import { ProviderEnvironment } from "@prosopo/env";
+import { type ProsopoConfigOutput, ScheduledTaskNames } from "@prosopo/types";
+import { CronJob } from "cron";
+import { Tasks } from "../tasks/tasks.js";
+import { checkIfTaskIsRunning } from "../util.js";
+
+export async function getClientList(
```
goastler, line 27: what does this do exactly?

packages/types/src/client/user.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export interface IUserData {
+	email: string;
+	name: string;
+	account: string;
+	url: string;
+	mnemonic: string;
+	createdAt: Timestamp;
+	activated: boolean;
+	tier?: Tier;
```
goastler: shouldn't a user always have a tier?

packages/types/src/client/user.ts
```diff
@@ -0,0 +1,50 @@
...
+	settings?: IUserSettings;
```
goastler: shouldn't a user always have settings?

packages/types/src/client/user.ts
```diff
@@ -0,0 +1,50 @@
...
+	updatedAtTimestamp?: Timestamp;
```
goastler: this should always be mandatory?

goastler (dismissed):

packages/database/src/databases/provider.ts
forgetso: you can pass the dbname in the URI so it's not needed here

packages/provider/src/schedulers/getClientList.ts
forgetso: Gets the client list from the client db on a schedule. It returns

```
{
    account
    settings
}[ ]
```

packages/types/src/client/user.ts
forgetso: No, because we don't set these at signup, they're added by the portal later. Zod creates defaults if there are no settings defined.

packages/types/src/client/user.ts
forgetso: Yes, changed.

packages/types/src/client/user.ts
forgetso: Which one? `tier` and `settings` were addressed previously.

`updatedAtTimestamp` doesn't exist initially because the record has not been updated, only created.

HughParry (approved):

PR #1355, opened 2024-09-12
Activated accounts only

forgetso:
- [x] stops site keys from using the service if they aren't registered


packages/common/src/locales/en.json
```diff
@@ -22,7 +22,8 @@
 		"INVALID_SIGNATURE": "Invalid signature",
 		"NOT_IMPLEMENTED": "Not implemented",
 		"SITE_KEY_MISSING": "SITE KEY missing",
-		"ACCOUNT_NOT_FOUND": "Account not found"
+		"ACCOUNT_NOT_FOUND": "Account not found",
+		"INVALID_TIMESTAMP": "Invalid timestamp"
```
goastler: do we have the translations added for the new strings?

packages/common/src/logger.ts
```diff
@@ -136,3 +136,9 @@ export class Loggable {
 		this.#logger = logger;
 	}
 }
+
+export const logError = (err: unknown, logger: Logger): void => {
+	logger.error(
+		typeof err === "object" && err ? ("stack" in err ? err.stack : err) : err,
```
goastler, line 142: if the type of the err is object, won't this output "false"?

packages/procaptcha/src/modules/ProsopoCaptchaApi.ts
```diff
@@ -60,6 +60,11 @@ export class ProsopoCaptchaApi implements ProsopoCaptchaApiInterface {
 				this.userAccount,
 				this.provider,
 			);
+
+			if (captchaChallenge[ApiParams.error]) {
```
goastler, line 64: why is this needed? It only skips the http->https code

packages/provider/src/api/captcha.ts
```diff
@@ -167,12 +175,28 @@ export function prosopoRouter(env: ProviderEnvironment): Router {
 		try {
 			const { user, dapp } = GetPowCaptchaChallengeRequestBody.parse(req.body);
 
+			validateAddress(user, false, 42);
+			validateAddress(dapp, false, 42);
```
goastler, line 191: duplicate?

goastler (changes requested):

packages/provider/src/api/captcha.ts
forgetso: No, one is validating the `dapp` and the other is validating the `user`

packages/procaptcha/src/modules/ProsopoCaptchaApi.ts
forgetso: because without the return the `captchas` key doesn't exist on the response so the following loop would fail. The API can now return a response with an `error` key, which, when encountered, is displayed in the widget as a red error message.

packages/common/src/logger.ts
forgetso: if the type of err is an object and err is not undefined
then if stack is in err, return err.stack
else return err
else return err

how does false come into this?

packages/common/src/locales/en.json
forgetso: Yes, they were already present but I needed to move them into the `GENERAL` section which I've now done.

forgetso: I've also added a unit test to catch any missing translation keys

forgetso: https://github.com/prosopo/captcha/blob/5c4ace994c7208e8b25c73473c5d16487a8ef94a/packages/common/src/tests/locales.unit.test.ts

forgetso: This stuff is already in due to the locale PR.

packages/locale/.npmignore
goastler, line 1: remove pls

goastler (dismissed):

packages/common/src/logger.ts
goastler:
```
typeof err === "object" && err ? ("stack" in err ? err.stack : err) : err
```
say err is not object, it becomes
```
false && err ? ("stack" in err ? err.stack : err) : err
```
false && <anything> is false, so
```
false
```

forgetso: I'm ok with this. Most likely scenario is that error will be `undefined` but there's a very small chance of that considering its in try catch block.

PR #1356, opened 2024-09-12
Deps

goastler:
went through and depcheck'd all pkgs


dev/config/package.json
```diff
@@ -31,58 +31,30 @@
 	"author": "Prosopo Limited",
 	"license": "Apache-2.0",
 	"dependencies": {
-		"@babel/core": "^7.24.5",
```
forgetso: I think all of these babel plugins are needed for the webpack config that's exported from config.

forgetso (changes requested):
we need to check the deps in config - probably best just to leave them alone.

dev/config/package.json
goastler: Okay so I moved all of these to procaptcha/bundle and then eliminated deps until it didn't build. Thinking about it though, the webpack etc deps should be in the config pkg bc that's where the webpack base config lives. That way any pkg that depends on config automatically inherits the correct deps for webpack.

I've reverted the eliminated deps to what they were based on I have no idea what they're doing / where they're used exactly other than a project search. I'm not too worried about the webpack deps bc they should be eliminated easily by vite as they're only used in one place: procaptcha-bundle.

The rollup deps I'm less sure on, and they'll be used with vite anytime it's bundling to they're defo a source of bloat. It's not too bad tho so can be left as is

Can you review procaptcha-bundle and config pkg deps and let me know if they're ok? they're the only ones that have changed since your comment

forgetso (approved):

PR #1367, opened 2024-09-16
replace npm ci with npm i for cache usage

goastler:


goastler:
not doing this anymore due to dependencies of dependencies getting bumped during npm i

PR #1391, opened 2024-09-23
Tech debt

HughParry:


packages/account/src/config/getPair.ts
```diff
@@ -73,24 +73,4 @@ export function getPair(
 	} else {
 		throw new ProsopoEnvError("GENERAL.NO_MNEMONIC_OR_SEED");
 	}
-}
-
-export function getReadOnlyPair(
```
goastler, line 78: where did this go?

packages/account/src/extension/ExtensionWeb2.ts
```diff
@@ -36,9 +36,8 @@ type AccountWithKeyPair = InjectedAccount & { keypair: KeyringPair };
  */
 export class ExtensionWeb2 extends Extension {
 	public async getAccount(
-		config: ProcaptchaClientConfigOutput,
 	): Promise<Account> {
-		const account = await this.createAccount(config);
+		const account = await this.createAccount();
```
goastler, line 40: do we not need a config for this anymore?

packages/account/tsconfig.json
```diff
@@ -7,7 +7,7 @@
 		"allowJs": true,
 		"jsxImportSource": "@emotion/react"
 	},
-	"include": ["src", "src/**/*.json"],
+	"include": ["src", "src/**/*.json", "../contract/src/accounts/getPair.ts"],
```
goastler: I thought the contracts pkg was removed?

packages/account/src/config/getMnemonic.ts
forgetso, line 1: Could this be changed to the same name as the function please?

packages/common/src/tests/array.test.ts
```diff
@@ -0,0 +1,26 @@
+import { describe, it, expect } from 'vitest'
+import { arrayJoin, ARRAY_JOINER } from '../array.js'
+
+describe('arrayJoin', () => {
```
goastler, line 4: probably better off using lodash from the utils pkg rather than rolling our own impl of arrayJoin

packages/account/tsconfig.json
forgetso: Is this a mistake? The package shouldn't reference a file in another package.

forgetso (changes requested):
2 small comments but overall LGTM

packages/common/src/tests/logger.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  })
+
+  it('should set log level correctly', () => {
+    const logger = getLogger('info', 'test')
+    logger.setLogLevel('warn')
+    expect(logger.getLogLevel()).toBe(LogLevel.enum.warn)
+  })
+
+  it('should throw for invalid log level', () => {
+    expect(() => getLogger('invalid' as any, 'test')).toThrow()
```
goastler: shouldn't need the `as any` here afaict

packages/account/tsconfig.json
HughParry: Removed

packages/account/src/config/getMnemonic.ts
HughParry: Resolved

packages/common/src/tests/logger.test.ts
goastler, line 1: where did the test go that tries setting each log level in turn?

this no longer tests setting the log level via the log level type itself rather than a string

goastler (changes requested):

packages/account/src/config/getPair.ts
HughParry: It's not used anywhere

packages/account/tsconfig.json
HughParry: Resolved

packages/common/src/tests/array.test.ts
HughParry: Sure but I've not added the arrayJoin function in here. If you want to remove then that's good with me

packages/common/src/tests/logger.test.ts
HughParry: This is a new test - I haven't removed any tests

packages/common/src/tests/logger.test.ts
HughParry: Good shout

packages/account/src/extension/ExtensionWeb2.ts
HughParry: No we don't

packages/common/src/tests/array.test.ts
goastler: will do in followup ticket

packages/common/src/tests/logger.test.ts
goastler: ah I see, mb :)

goastler (approved):

PR #1392, opened 2024-09-23
Better tsconfig checker

goastler:
- [x] examine pkg jsons against tsconfigs
- [x] flag any bad paths for references in tsconfig
- [x] flag any missing refs
- [x] flag any extra refs
- [x] add npm script


goastler:
wtf >300 file I only added 6 :roll_eyes:

PR #1393, opened 2024-09-23
Locale overhaul

forgetso:
- [x] separate locale from common package
- [x] separate locale into [JSON + backend] package and frontend package
- [x] add test to check if locale JSON strings are up to date


packages/database/src/databases/provider.ts
```diff
@@ -16,8 +16,10 @@ import {
 	type Logger,
 	ProsopoDBError,
 	ProsopoEnvError,
-	type TranslationKey,
 } from "@prosopo/common";
+importtype {
```
goastler: typo

packages/locale-browser/.npmignore
goastler, line 1: delete pls

packages/locale-browser/vite.test.config.ts
```diff
@@ -0,0 +1,32 @@
+import fs from "node:fs";
```
goastler: imports above license

packages/locale/.npmignore
goastler, line 1: remove pls

goastler (changes requested):

packages/database/src/databases/provider.ts
forgetso: Resolved. It was breaking the build.

packages/locale/.npmignore
forgetso: Removed

packages/locale-browser/vite.test.config.ts
forgetso: Sorted

packages/locale-browser/.npmignore
forgetso: Removed

goastler (approved):

PR #1423, opened 2024-10-01
Add button for restarting the captcha process

forgetso:
[Screencast from 01-10-24 21:09:49.webm](https://github.com/user-attachments/assets/6381ff67-830c-4bd6-b5cf-a98468e7b697)

TODO
- [x] button added to completely restart the image captcha process, losing all state
- [x] allow users to hook into onReload event
- [ ] limit the number of reloads in the client
- [ ] tell the provider that a request is a reload
- [ ] limit the number of reloads in the provider
- [ ] ensure the user doesn't get the same captcha again?


packages/procaptcha-pow/src/components/ProcaptchaWidget.tsx
```diff
@@ -46,30 +46,14 @@ const Procaptcha = (props: ProcaptchaProps) => {
...
-		if (!element) return;
-
-		const form = element.closest("form");
-		if (!form) return;
-
-		const handleSubmit = () => {
-			manager.current.resetState();
-		};
-
-		form.addEventListener("submit", handleSubmit);
```
goastler, line 64: why has this been removed?

goastler (changes requested):

packages/procaptcha-pow/src/components/ProcaptchaWidget.tsx
HughParry: See point 1 here https://prosopo.slack.com/archives/C039W3E3UNB/p1728596930956019

PR #1430, opened 2024-10-03
Make sure error message shows in image captcha widget

forgetso:
- [x] error messages in procaptcha widget
- [x] allow `onFailed` to be set by clients
- [x] allow bundle example to run all captcha tests, including the server verify parts
- [x] adds cypress test for selecting incorrect images and ensures that custom `onFailed` function is fired
- [x] restrict API admin auth middleware to prosopo API paths in provider API
- [x] use environment variables in bundle demo instead of hard coded site keys
- [x] add user and site key logging for tracing errors
- [x] add register site key function to setup script
- [x] add unit test for renderLogic to ensure that callbacks are picked up from data tags
- [x] add unit tests for errorHandler to make sure parsing errors are passed through
- [x] add integration tests to make sure new errors are passed back
- [x] changes image captcha process to use `timestamp` signature in line with pow process
- [x] removes defunct user verify API path
- [x] fixes bug with vite config where `defaultEnvironment` could be set to something other than [development, staging, production]
- [x] add `reset` function to procaptcha-budle for re-rendering the procaptcha widgets
- [x] add `onReset` callback
- [x] pass an error JSON object in API errors, using translation keys as codes
### Invalid Site Key
image
### Site Key Not Registered
image
### Failed to get Dataset
image


packages/procaptcha-bundle/src/index.tsx
```diff
@@ -114,26 +77,47 @@ export default function ready(fn: () => void) {
...
+	if (onloadUrlCallback) {
+		const onloadCallback = getWindowCallback(onloadUrlCallback);
+		// Add event listener to the script tag to call the callback function when the script is loaded
+		getProcaptchaScript(BUNDLE_NAME)?.addEventListener("load", () => {
+			ready(onloadCallback);
+		});
+		// or if the document has already loaded, call the callback function
+		if (document.readyState === "complete") {
+			ready(onloadCallback);
+		}
```
HughParry, line 109: Can this cause ready() to be called twice?

forgetso: Added a check.

HughParry (approved):

PR #1433, opened 2024-10-07
Mongo v7

goastler:


forgetso (changes requested):
This PR will need to modify the default image set in ansible vars as well.

PR #1434, opened 2024-10-07
Open observe / vector

goastler:


goastler:
blocked by #1433

docker/docker-compose.provider.yml
```diff
@@ -17,13 +17,21 @@ services:
             options:
                 max-size: '100m'
                 max-file: '1'
+        healthcheck:
+            test: ["CMD", "curl", "--fail", "localhost:9229/v1/prosopo/provider/status"]  # ping the status api
```
forgetso, line 21: Is this going to submit more notifications in the slack monitoring channel?

docker/docker-compose.provider.yml
```diff
@@ -142,6 +157,34 @@ services:
             options:
                 max-size: '100m'
                 max-file: '1'
+    vector:
```
forgetso, line 160: We'll need to see how staging copes with an additional container in terms of RAM / CPU when this is pushed to staging.

docker/images/vector/vector.toml
```diff
@@ -0,0 +1,188 @@
+# enable vector's api
```
forgetso, line 1: Well done for deciphering this format. Looks horrendous.

forgetso (approved):
I think we just need to trust you on this one George and hope that it runs when released into staging. The configs are over my head but the docker files look ok.

PR #1441, opened 2024-10-10
Log errors returned from provider and return false for verified

forgetso:


packages/types/src/provider/api.ts
```diff
@@ -234,7 +234,7 @@ export interface ProviderRegistered {
 
 export type ApiJsonError = {
 	message: string;
-	code: number;
+	code: string | number;
```
goastler: is the string a code in this situation? e.g. "500". If so, can we parseInt() to keep a single type

goastler (dismissed):

packages/types/src/provider/api.ts
HughParry: sorted

PR #1450, opened 2024-10-14
remove npm release check for testing provider mock

goastler:


goastler:
could do with @forgetso taking a look at this. Not sure if there was a reason why the provider mock was waiting on npm releases

HughParry (commented):
This reverts bugfix here https://github.com/prosopo/captcha/pull/1450

The dockerfile is small and only grabs the demo packages it needs so is only pulling from npm not from workspace

goastler:
closing for pr into staging

PR #1470, opened 2024-10-23
Stop captcha storage job from killing express

HughParry:


packages/provider/src/schedulers/captchaScheduler.ts
```diff
@@ -36,22 +33,28 @@ export async function storeCaptchasExternally(
...
-		const taskRunning = await checkIfTaskIsRunning(
-			ScheduledTaskNames.StoreCommitmentsExternal,
-			env.getDb(),
-		);
-		env.logger.info(
-			`${ScheduledTaskNames.StoreCommitmentsExternal} task running: ${taskRunning}`,
+	const job = new CronJob(cronSchedule, () => {
+		console.log(`Creating worker - from main thread: ${threadId}`);
+		const worker = new Worker(
+			path.resolve("../provider/dist/workers/storeCaptchaWorker.js"),
```
forgetso: Is it ok to have `dist` here?

packages/provider/src/schedulers/captchaScheduler.ts
```diff
@@ -36,22 +33,28 @@ export async function storeCaptchasExternally(
...
-			env.logger.info(
-				`${ScheduledTaskNames.StoreCommitmentsExternal} task....`,
-			);
-			await tasks.clientTaskManager.storeCommitmentsExternal().catch((err) => {
-				env.logger.error(err);
-			});
-		}
+
+		worker.on("message", (message) => {
+			console.log(`Worker message: ${message}`);
```
forgetso: Please use the logger for consistency. We'll be changing all of this to JSON output using the @prosopo/common logger soon.

packages/provider/src/schedulers/captchaScheduler.ts
```diff
@@ -12,21 +12,18 @@
...
 
 export async function storeCaptchasExternally(
-	pair: KeyringPair,
 	config: ProsopoConfigOutput,
 ) {
-	const env = new ProviderEnvironment(config, pair);
-	await env.isReady();
-
-	const tasks = new Tasks(env);
+	console.log(
```
forgetso: @prosopo/common logger

packages/provider/src/schedulers/captchaScheduler.ts
forgetso: Should we not be using `import.meta.module` or something or will this completely break in bundling? Actually, does this work after bundling CLI?

forgetso (changes requested):
Some comments regarding the path to worker file and nitpicking over logging.

forgetso (approved):

PR #1481, opened 2024-10-29
Working record batching

HughParry:


forgetso (approved):

packages/provider/src/tasks/client/clientTasks.ts
```diff
@@ -67,81 +67,69 @@ export class ClientTaskManager {
...
-				);
-			}
 			await this.providerDB.updateScheduledTaskStatus(
 				taskID,
 				ScheduledTaskStatus.Completed,
 				{
 					data: {
-						commitments: commitments.map((c) => c.id),
-						powRecords: powRecords.map((pr) => pr.challenge),
+						processedCommitments,
```
forgetso, line 131: Can you change this to a a count please?

HughParry: It is a count

PR #1497, opened 2024-11-04
Add domain restriction

HughParry:


packages/cli/src/start.ts
```diff
@@ -50,14 +52,18 @@
...
 	// Blocking middleware will run on any routes defined after this point
 	apiApp.use(blockMiddleware(env));
+
+	apiApp.use("/v1/prosopo/provider/client/", domainMiddleware(env));
 	apiApp.use(prosopoRouter(env));
 	apiApp.use(prosopoVerifyRouter(env));
 
 	if (admin) {
-		apiApp.use(authMiddleware(env));
+		apiApp.use("/v1/prosopo/provider/admin/", authMiddleware(env));
```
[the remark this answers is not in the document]
HughParry: copilot is telling lies

packages/cli/src/start.ts
```diff
@@ -50,14 +53,17 @@ function startApi(
 	apiApp.use(i18nMiddleware({}));
 	// Blocking middleware will run on any routes defined after this point
 	apiApp.use(blockMiddleware(env));
-	apiApp.use(prosopoRouter(env));
 	apiApp.use(prosopoVerifyRouter(env));
+	apiApp.use("/v1/prosopo/provider/client/", domainMiddleware(env));
```
forgetso, line 59: This should be an admin restricted endpoint as it triggers a scheduled task. It will need the concept of an `adminAccount` that is the same account we use on the lambdas. Use that account to authorise the call in the authMiddleware.ts

packages/provider/src/api/domainMiddleware.ts
```diff
@@ -0,0 +1,90 @@
...
+import type { NextFunction, Request, Response } from "express";
+import { Tasks } from "../tasks/index.js";
+import { handleErrors } from "./errorHandler.js";
+
+export const domainMiddleware = (env: ProviderEnvironment) => {
+	const tasks = new Tasks(env);
+
+	return async (req: Request, res: Response, next: NextFunction) => {
+		try {
+			const sitekeyInParams = req.params.dapp;
```
forgetso: Site key and user should be in headers for all of these routes. Use that as a first pass.

forgetso (changes requested):
Adding this now and will continue to review later.

dev/scripts/env.production
```diff
@@ -30,3 +30,6 @@ WATCHTOWER_NOTIFICATION_SLACK_HOOK_URL=
 WATCHTOWER_NOTIFICATIONS=
 WATCHTOWER_POLL_INTERVAL=30
 WATCHTOWER_REMOVE_VOLUMES=true
+PROSOPO_ADMIN_ADDRESS=5EjTA28bKSbFPPyMbUjNtArxyqjwq38r1BapVmLZShaqEedV
```
forgetso: Is this needed in production?

packages/cli/src/start.ts
```diff
@@ -88,7 +92,8 @@ export async function start(
 		});
 
 		const pair = await getPairAsync(secret);
-		env = new ProviderEnvironment(config, pair);
+		const authAccount = await getPairAsync(config.authAccount.secret);
```
forgetso: I don't think this should be secret for the providers cause that means they need to retain the secret in their env files. You can getPairAsync using an address only by doing 

```
getPairAsync( undefined, address )
```

and then you can use this pair to confirm signatures.

packages/provider/src/api/authMiddleware.ts
```diff
@@ -29,18 +29,18 @@ export const authMiddleware = (env: ProviderEnvironment) => {
 
 			const { signature, timestamp } = extractHeaders(req);
 
-			if (!env.pair) {
+			if (!env.authAccount) {
```
forgetso: What about if a provider calls their admin API endpoint with their own pair, `env.pair`? This should also be allowed.

packages/provider/src/api/domainMiddleware.ts
```diff
@@ -0,0 +1,86 @@
...
+				validateAddress(dapp, false, 42);
+			} catch (err) {
+				throw invalidSiteKeyError(dapp);
+			}
+
+			const clientSettings = await tasks.db.getClientRecord(dapp);
+			if (!clientSettings) throw siteKeyNotRegisteredError(dapp);
+
+			const allowedDomains = clientSettings.settings?.domains;
+			if (!allowedDomains) return next(siteKeyNotRegisteredError(dapp));
```
forgetso: Probably should be a different error like 

`siteKeyInvalidDomain`

packages/provider/src/api/domainMiddleware.ts
```diff
@@ -0,0 +1,86 @@
...
+
+			const origin = req.headers.origin;
+			if (!origin) throw siteKeyNotRegisteredError(dapp);
```
forgetso: `unauthorizedOriginError` ?

packages/provider/src/api/domainMiddleware.ts
```diff
@@ -0,0 +1,86 @@
...
+				if (tasks.clientTaskManager.isSubdomainOrExactMatch(origin, domain)) {
+					next();
+					return;
+				}
+			}
+
+			throw unauthorizedOriginError();
+		} catch (err) {
+			if (err instanceof ProsopoApiError) {
+				handleErrors(err, req, res, next);
```
forgetso, line 61: `handleErrors` can take ZodErrors as well I think

packages/provider/src/tasks/client/clientTasks.ts
```diff
@@ -249,6 +249,48 @@ export class ClientTaskManager {
 		await this.providerDB.storeUserBlockRuleRecords(rules);
 	}
 
+	isSubdomainOrExactMatch(referrer: string, clientDomain: string): boolean {
+		if (!referrer || !clientDomain) return false;
+		try {
+			const cleanReferrer = referrer.toLowerCase().trim().replace(/\/+$/, "");
```
forgetso: This should be fixed.

packages/provider/src/tasks/client/clientTasks.ts
```diff
@@ -249,6 +249,48 @@ export class ClientTaskManager {
...
+			const cleanAllowedInput = clientDomain
+				.toLowerCase()
+				.trim()
+				.replace(/\/+$/, "");
+
+			// Extract domain from URL if necessary
+			const getDomain = (url: string) => {
```
forgetso: Please replace or use https://github.com/prosopo/captcha/blob/ed6d1a72c1f839f2d94c45d74df3633c145f172e/packages/util/src/url.ts#L21-L40 instead. Could just return the new URL thing in a new function and keep `validateDomain` separate. The main requirement is having a single domain parser that we use across portal / lambdas / provider. This sort of thing should be in util.

packages/provider/src/tests/integration/imgCaptcha.integration.test.ts
```diff
@@ -41,10 +41,16 @@ describe("Image Captcha Integration Tests", () => {
 
 		it("should supply an image captcha challenge to a Dapp User", async () => {
 			const origin = "http://localhost";
-			const getImageCaptchaURL: TGetImageCaptchaChallengeURL = `${baseUrl}${ApiPaths.GetImageCaptchaChallenge}/${solutions.datasetId}/${userAccount}/${dappAccount}`;
+			const getImageCaptchaURL = `${baseUrl}${ApiPaths.GetImageCaptchaChallenge}`;
+			const getImgCaptchaBody = {
```
forgetso: Add type

packages/provider/src/tests/integration/imgCaptcha.integration.test.ts
```diff
@@ -59,10 +65,16 @@ describe("Image Captcha Integration Tests", () => {
 		it("should not supply an image captcha challenge to a Dapp User if the site key is not registered", async () => {
 			const origin = "http://localhost";
 			const [_mnemonic, unregisteredAccount] = await generateMnemonic();
-			const getImageCaptchaURL: TGetImageCaptchaChallengeURL = `${baseUrl}${ApiPaths.GetImageCaptchaChallenge}/${solutions.datasetId}/${userAccount}/${unregisteredAccount}`;
+			const getImageCaptchaURL = `${baseUrl}${ApiPaths.GetImageCaptchaChallenge}`;
+			const body = {
```
forgetso: Add type

packages/provider/src/tests/integration/imgCaptcha.integration.test.ts
```diff
@@ -96,11 +114,19 @@ describe("Image Captcha Integration Tests", () => {
 
 		it("should fail if datasetID is incorrect", async () => {
 			const datasetId = "thewrongdsetId";
-			const getImageCaptchaURL: TGetImageCaptchaChallengeURL = `${baseUrl}${ApiPaths.GetImageCaptchaChallenge}/${datasetId}/${userAccount}/${dappAccount}`;
+			const origin = "http://localhost";
+			const getImageCaptchaURL = `${baseUrl}${ApiPaths.GetImageCaptchaChallenge}`;
+			const body = {
```
forgetso: Add type

packages/provider/src/tests/integration/imgCaptcha.integration.test.ts
```diff
@@ -78,10 +90,16 @@ describe("Image Captcha Integration Tests", () => {
 		it("should not supply an image captcha challenge to a Dapp User if an invalid site key is provided", async () => {
 			const invalidSiteKey = "junk";
 			const origin = "http://localhost";
-			const getImageCaptchaURL: TGetImageCaptchaChallengeURL = `${baseUrl}${ApiPaths.GetImageCaptchaChallenge}/${solutions.datasetId}/${userAccount}/${invalidSiteKey}`;
+			const getImageCaptchaURL = `${baseUrl}${ApiPaths.GetImageCaptchaChallenge}`;
+			const body = {
```
forgetso: Add type

packages/provider/src/tests/integration/imgCaptcha.integration.test.ts
```diff
@@ -118,9 +144,15 @@ describe("Image Captcha Integration Tests", () => {
 
 			const userAccount = dummyUserAccount.address;
 			const origin = "http://localhost";
-			const getImageCaptchaURL: TGetImageCaptchaChallengeURL = `${baseUrl}${ApiPaths.GetImageCaptchaChallenge}/${solutions.datasetId}/${userAccount}/${dappAccount}`;
+			const getImageCaptchaURL = `${baseUrl}${ApiPaths.GetImageCaptchaChallenge}`;
+			const getImgCaptchaBody = {
```
forgetso: Add type

packages/provider/src/tasks/client/clientTasks.ts
```diff
@@ -249,6 +249,48 @@ export class ClientTaskManager {
...
+						return new URL(url).hostname;
+					}
+					// Otherwise, clean up any paths or ports
+					const parts = url.split("/");
+					if (!parts[0]) return url;
+					const hostParts = parts[0].split(":");
+					if (!hostParts[0]) return url;
+					return hostParts[0];
+				} catch {
+					return url;
```
forgetso: We should at least log these

packages/provider/src/tasks/client/clientTasks.ts
```diff
@@ -249,6 +249,48 @@ export class ClientTaskManager {
...
+				try {
+					// If it's a full URL, use URL parser
+					if (url.includes("://")) {
+						return new URL(url).hostname;
+					}
+					// Otherwise, clean up any paths or ports
+					const parts = url.split("/");
+					if (!parts[0]) return url;
+					const hostParts = parts[0].split(":");
+					if (!hostParts[0]) return url;
```
forgetso: We should at least log these

packages/provider/src/tasks/client/clientTasks.ts
```diff
@@ -249,6 +249,48 @@ export class ClientTaskManager {
...
+			// Extract domain from URL if necessary
+			const getDomain = (url: string) => {
+				try {
+					// If it's a full URL, use URL parser
+					if (url.includes("://")) {
+						return new URL(url).hostname;
+					}
+					// Otherwise, clean up any paths or ports
+					const parts = url.split("/");
+					if (!parts[0]) return url;
```
forgetso: We should at least log these

packages/provider/src/tests/unit/tasks/client/clientTasks.unit.test.ts
```diff
@@ -371,4 +371,153 @@ describe("ClientTaskManager", () => {
...
+				),
+			).toBe(true);
+			expect(
+				clientTaskManager.isSubdomainOrExactMatch(
+					"example.com",
+					"https://example.com:3000",
+				),
+			).toBe(true);
+			expect(
+				clientTaskManager.isSubdomainOrExactMatch(
```
forgetso, line 516: Does this work for
```
⭐⭐⭐⭐.com
漢字漢字漢字.com
+ Other weird styles of domain
```

packages/types/src/config/config.ts
```diff
@@ -295,6 +295,11 @@ export const ProsopoConfigSchema = ProsopoBasicConfigSchema.merge(
 		rateLimits: ApiPathRateLimits.default(ProviderDefaultRateLimits),
 		proxyCount: number().optional().default(0),
 		lRules: record(string(), number()).optional(),
+		authAccount: object({
+			address: string().optional(),
+			secret: string().optional(),
```
forgetso, line 300: We probably only need / want `address`.

forgetso (changes requested):
Looks good overall but some points on secret management and domain checker.

packages/types/src/config/config.ts
HughParry: Sure but this is also used in the cli, where we would need the full acc to sign stuff

packages/provider/src/api/domainMiddleware.ts
HughParry: Yeah wasn't sure what error for no origin - I almost feel like telling them why it's failed does nothing but give an attacker more info

packages/provider/src/api/authMiddleware.ts
HughParry: Just to be clear - we want to attempt to verify with both the authAccount and the provider pair? For all routes?

packages/provider/src/tests/unit/tasks/client/clientTasks.unit.test.ts
HughParry: yeah works fine. Added some stupidly weird test cases

packages/provider/src/api/domainMiddleware.ts
HughParry: added

dev/scripts/env.production
HughParry: not sure, setup of e2es in gha is difficult to replicate locally. Will drop and see if passing

packages/cli/src/start.ts
HughParry: resolved

packages/cli/src/start.ts
HughParry: done

packages/provider/src/api/authMiddleware.ts
HughParry: done

packages/provider/src/api/domainMiddleware.ts
HughParry: added to unauthorizedOriginError for calls that haven't supplied origin

packages/provider/src/tasks/client/clientTasks.ts
HughParry: done

packages/provider/src/tasks/client/clientTasks.ts
HughParry: resolved

packages/provider/src/tasks/client/clientTasks.ts
HughParry: resolved

forgetso (approved):

forgetso (approved):

PR #1508, opened 2024-11-11
validate with wildcard *

HughParry:


forgetso (approved):

forgetso (changes requested):
Put this business logic in the lambdas and portal page?

PR #1555, opened 2024-12-03
zod error type + parsing fn

goastler:


packages/common/src/zod.ts
```diff
@@ -0,0 +1,57 @@
...
+	throw new ProsopoZodParseError(`\n${msg}`, {
+		context: {
+			zodErrors: result.error.errors,
+		},
+	});
+};
+
+export type ZodParseParams = BaseContextParams & { zodErrors?: z.ZodIssue[] };
+
+export class ProsopoZodParseError extends ProsopoBaseError<ZodParseParams> {
```
HughParry: Can we define the error class in error.ts so that they're all in the same place

packages/common/src/error.ts
```diff
@@ -24,18 +25,20 @@ type BaseErrorOptions<ContextType> = {
 	silent?: boolean;
 };
 
-interface BaseContextParams {
+export interface BaseContextParams {
 	// biome-ignore lint/suspicious/noExplicitAny: TODO remove any
 	[key: string]: any;
```
HughParry, line 30: while we're here can we change this to 

[key: string]: unknown;

and then drop the biome-ignore

HughParry (changes requested):
nitpicking

packages/common/src/zod.ts
goastler: done

packages/common/src/error.ts
goastler: change to what?

HughParry: so weird, didn't post properly

[key: string]: unknown;

goastler: done

PR #1558, opened 2024-12-04
Move the error unwrapper into common

forgetso:


dev/config/src/vite/vite.test.config.ts
```diff
@@ -52,6 +53,7 @@ export default function () {
 			},
 			pool: "forks", // forks is slower than 'threads' but more compatible with low-level libs (e.g. bcrypt)
 			testTimeout: 10000,
+			//disableConsoleIntercept: true,
```
HughParry: Remove new unused comment

packages/common/src/error.ts
```diff
@@ -176,3 +181,42 @@ export class ProsopoApiError extends ProsopoBaseError<ApiContextParams> {
...
+		} else {
+			jsonError.message = JSON.parse(err.message);
+		}
+	}
+
+	jsonError.code = jsonError.code || code;
+	return { code, statusMessage, jsonError };
+};
+
+export function isZodError(err: unknown): err is ZodError {
```
HughParry: using consts or functions? different declaration in same file between line 184 and line 218

packages/common/src/error.ts
```diff
@@ -176,3 +181,42 @@ export class ProsopoApiError extends ProsopoBaseError<ApiContextParams> {
 		this.code = code;
 	}
 }
+
+export const unwrapError = (err: ProsopoApiError | SyntaxError | ZodError) => {
```
HughParry: Why is this typed as a ProsopoApiError? Aren't there other error types which we might want to unwrap?

HughParry (approved):

packages/common/src/error.ts
forgetso: I've changed it to a const.

Don't understand the second comment. What do you mean about a different declaration?

packages/common/src/error.ts
forgetso: At the moment its only being used for ProsopoApiError, but yes it should probably be ProsopoBaseError. Changed.

packages/common/src/error.ts
HughParry: was trying to communicate:

Don't mind function or const, both are fine. Just imo best to keep consistent within files, not use both

PR #1564, opened 2024-12-09
ios webview fix

goastler:
- **remove problematic padding, fix overlay layout, use button tag for img containers**
- **adjust layout based on fixed img grid padding**


goastler:
DO NOT MERGE UNTIL CHECKED!

PR #1569, opened 2024-12-10
Simplifying outgoing links

HughParry:


packages/web-components/src/WidgetConstants.ts
```diff
@@ -11,8 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-export const WIDGET_URL =
-	"https://prosopo.io/?ref=prosopo.io&amp;utm_campaign=widget&amp;utm_medium=checkbox#features";
```
goastler, line 15: utm campaign gets dropped - is this on purpose?

goastler (changes requested):

packages/web-components/src/WidgetConstants.ts
HughParry: yeah

goastler (approved):

PR #1570, opened 2024-12-11
Ttl on sessions and reject old frictionless tokens

HughParry:


packages/provider/src/api/captcha.ts
```diff
@@ -42,6 +42,7 @@ import { getIPAddress } from "../util.js";
 import { handleErrors } from "./errorHandler.js";
 
 const DEFAULT_FRICTIONLESS_THRESHOLD = 0.5;
+const TEN_MINUTES = 60 * 10;
```
forgetso, line 45: Isn't this 600ms?

packages/types-database/src/types/provider.ts
```diff
@@ -59,6 +59,7 @@ export type IUserDataSlim = Pick<IUserData, "account" | "settings">;
 
 export type ClientRecord = IUserDataSlim & Document;
 
+const TEN_MINUTES = 60 * 10;
 const ONE_DAY = 60 * 60 * 24;
```
forgetso, line 63: Maybe these were wrong all along?

forgetso (changes requested):
Some small comments. Good investigation.

packages/provider/src/api/captcha.ts
forgetso: Mongoose is in seconds

forgetso (approved):

PR #1664, opened 2025-01-30
Frictionless retry

forgetso:
- [x] makes frictionles retry if it fails
- [x] passes a restart function from frictionless into image/pow components so that entry point is always frictioness session creation
- [x] unifies events/callbacks between bundle and components
- [x] standardises testing between client-example-bundle and client-example
- [x] adds cypress config for testing frictionless


packages/procaptcha-common/src/elements/window.ts
```diff
@@ -0,0 +1,24 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+export const getWindowCallback = (callbackName: string) => {
+	// biome-ignore lint/suspicious/noExplicitAny: TODO fix any
+	const fn = (window as any)[callbackName.replace("window.", "")];
```
HughParry, line 17: window as Record<string, unknown> would make biome happy and better reflects what's going on here

packages/procaptcha-frictionless/src/ProcaptchaFrictionless.tsx
```diff
@@ -12,82 +12,109 @@
...
-		const configOutput = ProcaptchaConfigSchema.parse(config);
+	const resetState = (attemptCount?: number) => {
+		stateRef.current = defaultLoadingState(
+			attemptCount || stateRef.current.attemptCount,
+		);
+	};
 
-		const detectAndSetComponent = async () => {
-			try {
+	const fallOverWithStyle = (errorMessage?: string) => {
```
HughParry, line 75: love u kept this. was thinking of this when naming

image

packages/procaptcha-frictionless/src/ProcaptchaFrictionless.tsx
```diff
@@ -106,13 +133,29 @@ export const ProcaptchaFrictionless = ({
...
+			5,
+		).finally(() => {
+			if (stateRef.current.attemptCount >= 5) {
+				fallOverWithStyle();
+				restartComponentTimeout();
 			}
+		});
+	};
+
+	// biome-ignore lint/correctness/useExhaustiveDependencies: <explanation>
```
HughParry, line 155: <explanation>

HughParry (approved):

PR #1692, opened 2025-02-26
Feature widget optimization

light-source:


demos/cypress-shared/cypress/e2e/frictionless.cy.ts
```diff
@@ -65,7 +65,7 @@ describe("Captchas", () => {
 					expect(response?.body).to.have.property("error");
 				})
 				.then(() => {
-					cy.get('div[data-cy="button-human"]', { includeShadowDom: true })
+					cy.get(".captcha-checkbox", { includeShadowDom: true })
```
forgetso: The idea behind having dynamic classes is that they are slightly harder to automate. If you stick a class on called `.captcha-checkbox`, it makes it arbitrary. `data-cy` tags are stripped when bundling. Is this class necessary?

demos/cypress-shared/cypress/e2e/frictionless.cy.ts
```diff
@@ -99,6 +99,11 @@ describe("Captchas", () => {
 	});
 
 	it("Captchas load when 'I am human' is pressed", () => {
+		cy.intercept("POST", "**/prosopo/provider/client/captcha/frictionless").as(
+			"frictionless",
+		);
+		cy.wait("@frictionless", { timeout: 36000 });
```
forgetso, line 105: Why so long?

dev/config/src/webpack/webpack.config.ts
```diff
@@ -12,13 +11,26 @@ import path from "node:path";
...
+import fs from "node:fs";
+import path from "node:path";
 import HtmlWebpackPlugin from "html-webpack-plugin";
 import MiniCssExtractPlugin from "mini-css-extract-plugin";
 import NodePolyfillPlugin from "node-polyfill-webpack-plugin";
 import TerserPlugin from "terser-webpack-plugin";
 
 const moduleDirs = [path.resolve("./node_modules")];
 
+function resolveEntryFile(): string {
```
forgetso, line 24: Why is this needed?

packages/database/src/databases/provider.ts
```diff
@@ -851,20 +851,7 @@ export class ProviderDatabase
...
-										},
-									},
-									{
-										$convert: {
-											input: "$lastUpdatedTimestamp",
-											to: "date",
-										},
-									},
-								],
+								$lt: ["$storedAtTimestamp", "$lastUpdatedTimestamp"],
```
forgetso: Please don't change this comparison.

packages/types/src/client/user.ts
```diff
@@ -48,15 +48,7 @@ export interface IUserData {
 	originUrl: string;
 	settings?: IUserSettings;
 	updatedAtTimestamp?: Timestamp;
-	tierCheckoutId?: string; // prosopo (not stripe!) checkout session id / lock
```
forgetso: Please don't remove these. I think you may need to merge staging.

forgetso (changes requested):

demos/cypress-shared/cypress/e2e/frictionless.cy.ts
light-source: I've updated this to `data-cy`. 

FYI: atm the `data-cy` is present in the Live captcha output - e.g. check the preview inside the WordPress plugin. Also, search inside the project doesn't show any rules or code against 'data-cy'.

demos/cypress-shared/cypress/e2e/frictionless.cy.ts
light-source: I've used the same timeout as similar code in the same file, e.g.

```
it("An error is returned if captcha type is set to image and frictionless is used in the widget
...
cy.wait("@getCaptcha", { timeout: 36000 })
```

Let me know if I should change this, and maybe others to something else.

dev/config/src/webpack/webpack.config.ts
light-source: the `packages/procaptcha-bundle` package had `index.tsx`, but this extension became confusing since the file does not contain any React elements anymore. I've changed to '.ts', and had to improve the webpack config, which was setup to 'index.tsx' entries only.

packages/database/src/databases/provider.ts
light-source: My apologies, it seems I've overlooked this while merging your changes.

packages/types/src/client/user.ts
light-source: My bad again, sorry. I had no changes in this package, and during merge conflict resolution I was accepting your version in all such cases. Maybe this case was resolved automatically by IDE using some file timestamp after formatting, IDK. 

I'll be even more careful next time.

demos/cypress-shared/cypress/e2e/frictionless.cy.ts
forgetso: Sure, np

demos/cypress-shared/cypress/e2e/frictionless.cy.ts
forgetso: Hmm, that's strange. I thought it was removed in the bundle. See from our live site:

```
<input name="XRwOhYua" id="XRwOhYua" type="checkbox" aria-live="assertive" aria-label="I am human" class="procaptcha-1bnw6ba" style="width: 28px; height: 28px; min-width: 14px; min-height: 14px; top: auto; left: auto; opacity: 1; border-radius: 12.5%; appearance: none; cursor: pointer; margin: 15px; border: 1px solid rgb(189, 189, 189); flex: 1 1 0%;">
```

forgetso: https://prosopo.io/contact/

light-source: `<div style="max-width: 302px; min-height: 80px;" data-cy="button-human">` I still see it in the contact captcha

forgetso: My bad, could you please stick a condition on this attribute so that it only appears when `NODE_ENV !== production`?

PR #1696, opened 2025-03-03
Zod native enum bug workaround

goastler:
Zod native enum has an issue with the enum values being typed correctly. This works around the issue until I find a fix. These are the only two nativeEnums used in the shared types pkg I'm working on for the portal + api, so this is the mvp workaround


goastler:
moot, think I found a solution

PR #1697, opened 2025-03-03
Client hello

forgetso:


demos/provider-mock/src/api.ts
```diff
@@ -85,5 +88,153 @@ export function prosopoRouter(): Router {
...
+			!headers["x-tls-clienthello"] ||
+			!headers["x-tls-version"] ||
+			!headers["x-tls-server-name"]
+		) {
+			console.error("Missing required headers.");
+			return res.status(400).send("Missing required headers.");
+		}
+
+		// make sure the client hello message is a string
+		if (
```
goastler, line 109: aren't headers always string-string key/value maps?

demos/provider-mock/src/api.ts
```diff
@@ -85,5 +88,153 @@ export function prosopoRouter(): Router {
...
+	router.get("/test", async (req, res) => {
+		console.log(req.headers);
+		console.log(req.rawHeaders);
+
+		// Extract headers
+		const headers = req.headers;
+
+		// Validate headers
+		if (
+			!headers["x-tls-clienthello"] ||
```
goastler, line 100: should we just set the headers to '' rather than telling the caller they've not been set?

demos/provider-mock/src/api.ts
```diff
@@ -85,5 +88,153 @@ export function prosopoRouter(): Router {
...
+
+			// Debug: Check first few bytes
+			console.log(
+				"ClientHello First Bytes:",
+				clientHelloBuffer.slice(0, 5).toString("hex"),
+			);
+
+			// Check first byte after the initial 5
+			if (clientHelloBuffer[5] !== 0x01) {
+				return { error: "Invalid ClientHello message: First byte is not 0x01" };
```
goastler, line 133: same deal here, might be better off not telling the caller what's going on

packages/database/src/databases/client.ts
```diff
@@ -25,9 +26,9 @@ import { MongoDatabase } from "../base/index.js";
 
 const CLIENT_TABLES = [
 	{
-		collectionName: TableNames.emails,
-		modelName: "Email",
-		schema: UserDataSchema,
+		collectionName: TableNames.accounts,
+		modelName: "Account",
```
goastler, line 30: is this not going to move all the data?

packages/locale/src/i18nMiddleware.ts
```diff
@@ -12,14 +12,15 @@
...
+import type { RequestHandler } from "express";
 import { type HandleOptions, handle } from "i18next-http-middleware";
 import loadI18next from "./loadI18next.js";
 
 async function i18nMiddleware(
 	options: HandleOptions,
 ): Promise<ReturnType<typeof handle>> {
 	const i18n = await loadI18next(true);
-	return handle(i18n, { ...options });
+	return handle(i18n, { ...options }) as RequestHandler;
```
goastler: is this needed? I thought this was resolved with an npm i

packages/provider/src/api/ja4.ts
```diff
@@ -0,0 +1,156 @@
...
+		!headers["x-tls-version"] ||
+		!headers["x-tls-server-name"]
+	) {
+		logger.error("Missing required headers.");
+		return { error: "Missing required headers." };
+	}
+
+	// make sure the client hello message is a string
+	if (
+		typeof headers["x-tls-clienthello"] !== "string" ||
```
goastler: see comments on provider mock version

packages/types-database/src/types/client.ts
```diff
@@ -43,8 +43,74 @@ export const UserDataSchema: mongoose.Schema<UserDataRecord> = new Schema({
 	updatedAtTimestamp: Number,
 });
 
+type User = {
```
goastler, line 46: do these need to be in here? wrt types, everything should be in private by default unless needed by the provider imo

goastler (dismissed):

demos/provider-mock/src/api.ts
forgetso: they are `string` or `string[]`

demos/provider-mock/src/api.ts
forgetso: yeah, probably, and then do something with null JA4 hashes later.

packages/database/src/databases/client.ts
forgetso: the data is already moved. This code is no longer functional because emails was migrated to accounts

packages/locale/src/i18nMiddleware.ts
forgetso: correct, removed.

packages/provider/src/api/ja4.ts
forgetso: I've fixe this in the provider code. Provider-mock API stuff is just for us testing without having to run a provider.

packages/types-database/src/types/client.ts
forgetso: I could remove the client getter job and then these wouldn't be needed. We'd be relying on the API `registerSiteKey` stuff always working in this case. Shall I remove?

demos/provider-mock/src/api.ts
forgetso: set to default empty strings if not present

HughParry (approved):

PR #1715, opened 2025-03-11
vector config

goastler:
- **enable healthchecks**
- **add batch config to caddy sink**
- **don't print caddy config**


goastler:
this is in the 2.5.1 rel

PR #1740, opened 2025-03-24
Feature: Captcha invisible mode

HughParry:


packages/procaptcha-bundle/src/util/widgetFactory.ts
```diff
@@ -54,8 +56,20 @@ class WidgetFactory {
 		const widgetTheme =
 			"light" === renderOptions.theme ? lightTheme : darkTheme;
 
-		const widgetInteractiveArea =
-			this.widgetSkeletonFactory.createWidgetSkeleton(container, widgetTheme);
+		let widgetInteractiveArea: HTMLElement;
+
+		// Don't create the widget skeleton if the mode is invisible
+		if (!invisible) {
```
forgetso: I'd probably prefer `if (invisible)` instead of `if not`

packages/procaptcha-frictionless/src/ProcaptchaFrictionless.tsx
```diff
@@ -31,11 +33,16 @@ import customDetectBot from "./customDetectBot.js";
 
 const renderPlaceholder = (
 	theme: string | undefined,
+	mode: ModeType,
 	errorMessage: string | undefined,
 ) => {
 	const { t } = useTranslation();
 	const checkboxTheme = "light" === theme ? lightTheme : darkTheme;
 
+	if (mode === "invisible") {
```
forgetso, line 42: enum or string literal would be better for `invisible`

packages/procaptcha-pow/src/components/ProcaptchaWidget.tsx
```diff
@@ -41,6 +44,41 @@ const Procaptcha = (props: ProcaptchaProps) => {
 		}
 	}, [config.language]);
 
+	// Add event listener for the execute event (works for invisible mode)
+	useEffect(() => {
+		// Only set up event listener if in invisible mode
+		if (config.mode === "invisible") {
```
forgetso: enum / type

forgetso: since you're using Zod you can use `Mode.Values.invisible` or something like that

forgetso (changes requested):
Small changes

PR #1746, opened 2025-03-25
Env per captcha type in demos

HughParry:


HughParry:
Closing since we have this in prod already

PR #1754, opened 2025-03-26
Feature framework integrations

light-source:


dev/lint/src/refs.ts
```diff
@@ -23,183 +24,220 @@ import z from "zod";
 type File = { path: string; content: any };
 
 // biome-ignore lint/suspicious/noExplicitAny: <explanation>
-const readJson = (filePath: string): any => {
+const parseJsonFile = (filePath: string): any => {
```
HughParry: Since we're refactoring!

1. Remove brackets and return from arrow funcs that immediately return
2. Change any for something else (probably unknown instead) and remove the biome ignore

HughParry: Same goes for other funcs in this file with the same issues

integration/frameworks/angular/angular-procaptcha-wrapper/package.json
```diff
@@ -0,0 +1,41 @@
...
+		}
+	},
+	"scripts": {
+		"test": "echo \"No test specified\"",
+		"clean": "del-cli ./dist",
+		"build": "ng-packagr -p ng-package.json",
+		"dev": "npm run build -- --watch"
+	},
+	"dependencies": {
+		"@prosopo/procaptcha-wrapper": ">=2.5.4"
```
HughParry, line 25: deps should be pinned. They get bumped by the github actions bump action

integration/frameworks/angular/angular-procaptcha-wrapper/package.json
```diff
@@ -0,0 +1,41 @@
+{
+	"name": "@prosopo/angular-procaptcha-wrapper",
+	"version": "1.0.0",
```
HughParry, line 3: As with other package.jsons, keep the versions in line with current versions for packages. This makes bumps and releases easier

integration/frameworks/react/react-procaptcha-integration-demo/src/app.tsx
```diff
@@ -0,0 +1,42 @@
...
+		}
+
+		throw new Error("root element is not found");
+	}
+
+	protected renderProcaptcha(rootElement: Element): void {
+		const reactRoot = ReactDOM.createRoot(rootElement);
+
+		reactRoot.render(
+			<React.StrictMode>
```
HughParry: If we end up using this for e2e tests then we need to make sure that it's run in production, since strictmode will cause double renders which won't play nicely with cypress (unless accounted for of course!)

integration/frameworks/react/react-procaptcha-integration-demo/src/integrationDemo.tsx
```diff
@@ -0,0 +1,41 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { ProcaptchaComponent } from "@prosopo/react-procaptcha-wrapper";
+import React from "react";
+
+class IntegrationDemo extends React.Component {
```
HughParry: Could we use react functional components here instead of classes please

integration/frameworks/react/react-procaptcha-wrapper/src/procaptchaComponent.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+type ProcaptchaComponentProperties = ProcaptchaRenderOptions & {
+	htmlAttributes: React.DetailedHTMLProps<
+		React.HTMLAttributes<HTMLDivElement>,
+		HTMLDivElement
+	>;
+};
+
+class ProcaptchaComponent extends React.Component<ProcaptchaComponentProperties> {
+	private readonly elementRef: React.RefObject<HTMLDivElement>;
```
HughParry: Use createRef instead

integration/frameworks/react/react-procaptcha-wrapper/src/procaptchaComponent.tsx
```diff
@@ -0,0 +1,58 @@
...
+import React from "react";
+
+type ProcaptchaComponentProperties = ProcaptchaRenderOptions & {
+	htmlAttributes: React.DetailedHTMLProps<
+		React.HTMLAttributes<HTMLDivElement>,
+		HTMLDivElement
+	>;
+};
+
+class ProcaptchaComponent extends React.Component<ProcaptchaComponentProperties> {
```
HughParry: Should be react functional components instead

integration/frameworks/react/react-procaptcha-wrapper/src/procaptchaComponent.tsx
```diff
@@ -0,0 +1,58 @@
...
+	private readonly elementRef: React.RefObject<HTMLDivElement>;
+
+	constructor(properties: ProcaptchaComponentProperties) {
+		super(properties);
+
+		this.elementRef = React.createRef();
+	}
+
+	public override componentDidMount(): void {
```
HughParry: we can just use useEffect hooks here instead

integration/frameworks/react/react-procaptcha-wrapper/src/procaptchaComponent.tsx
```diff
@@ -0,0 +1,58 @@
...
+		this.renderProcaptcha();
+	}
+
+	public override render(): React.ReactNode {
+		const { htmlAttributes = {} } = this.props;
+
+		return <div ref={this.elementRef} {...htmlAttributes} />;
+	}
+
+	protected async renderProcaptcha(): Promise<void> {
```
HughParry: Could this component instead look something like:

```
function ProcaptchaComponent(props) {
  const { htmlAttributes } = props
  const elementRef = useRef(null)

  useEffect(() => {
    const renderCaptcha = async () => {
        await renderProcaptcha(elementRef.current, props);
    };
    
    renderCaptcha()
  }, [props])

  return <div ref={elementRef} {...htmlAttributes} />
}
```

package.json
```diff
@@ -55,7 +55,7 @@
 		"docker:rust": "f() { if [ \"$CI\" = \"true\" ]; then FLAGS=''; else FLAGS='-it'; fi ; mkdir -p node_modules/cargo-cache && docker run --rm $FLAGS -u $(id -u):$(id -g) --cpu-quota=-1 -v $PWD:/src -v $PWD/node_modules/cargo-cache:/usr/local/cargo/registry prosopo/cargo-contract:4.1.1 \"export RUST_BACKTRACE=1; $@\" ;}; f",
 		"lint-changed:js": "npm run --silent lint:js -- --changed",
 		"lint-changed-fix:js": "npm run --silent lint-fix:js -- --changed",
-		"lint:version": "npm run -w @prosopo/lint version -- $(pwd)/package.json",
+		"lint:version": "npm run -w @prosopo/lint version -- --file=$(pwd)/package.json --exclude='integration/'",
```
HughParry, line 58: By pinning to name versions you can drop this change

package.json
```diff
@@ -95,6 +101,7 @@
 		"depcheck": "1.4.7",
 		"htmlhint": "1.1.4",
 		"husky": "9.1.7",
+		"lerna": "^8.2.1",
```
HughParry: pin deps

packages/locale/package.json
```diff
@@ -29,7 +29,7 @@
 		"i18next-chained-backend": "4.6.2",
 		"i18next-fs-backend": "2.6.0",
 		"i18next-http-backend": "3.0.2",
-		"i18next-http-middleware": "3.7.1",
+		"i18next-http-middleware": "3.6.0",
```
HughParry, line 32: chat to Chris on this one

packages/procaptcha-bundle/src/util/configCreator.ts
```diff
@@ -22,7 +22,7 @@ import {
 function createConfig(
 	siteKey?: string,
 	theme: "light" | "dark" = "light",
-	language?: typeof LanguageSchema,
+	language?: (typeof Languages)[number],
```
HughParry: Why?

packages/procaptcha-wrapper/src/index.ts
```diff
@@ -0,0 +1,50 @@
...
+import type { Languages } from "@prosopo/locale";
+import type { CaptchaType, ProcaptchaRenderOptions } from "@prosopo/types";
+import {
+	type RenderProcaptchaFunction,
+	loadRenderProcaptchaScript,
+} from "./renderProcaptcha.js";
+
+let renderFunction: RenderProcaptchaFunction;
+
+export const renderProcaptcha = async (
```
HughParry: Define renderProcaptcha elsewhere and export it from index.ts please

packages/procaptcha-wrapper/vite.config.ts
```diff
@@ -0,0 +1,29 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { createIntegrationViteConfig } from "@prosopo/procaptcha-integration-build-config";
+
+export default createIntegrationViteConfig({
```
HughParry, line 17: Might be easier to extend the vite frontend config here

HughParry (changes requested):

packages/procaptcha-bundle/src/util/configCreator.ts
light-source: because I needed to get rid of Zod in the [schema definition](https://github.com/prosopo/captcha/blob/17803a75556dae239289740cf9e312ebab05f69f/packages/types/src/procaptcha-bundle/index.ts), to make it portable.
In this case, I just repeated the same for compatibility.

integration/frameworks/angular/angular-procaptcha-wrapper/package.json
light-source: That's an intended one. In the current setup, procaptcha-wrapper's version will be in sync with the project one, while integrations will get new versions only when they are updated. 

I've used `>`, so the customers keep using the latest `renderOptions` when your project gets version 3.x and later.

Have you seen [the readme](https://github.com/prosopo/captcha/blob/17803a75556dae239289740cf9e312ebab05f69f/integration/readme.md) with the approach explanation?

integration/frameworks/angular/angular-procaptcha-wrapper/package.json
light-source: That's an intended one. 

Please check [the readme](https://github.com/prosopo/captcha/blob/17803a75556dae239289740cf9e312ebab05f69f/integration/readme.md) with the approach explanation

packages/procaptcha-wrapper/vite.config.ts
light-source: This case requires `vite-plugin-dts` usage and specific settings for it, which is missing in the `dev/config` package.

Reusing integration config seems better imho.

packages/locale/package.json
light-source: @forgetso Is it okay for you to have `i18next-http-middleware` as `3.6.0` instead of `3.7.1`?

As I mentioned in Slack, the reason for this change is matching the project's `@types/express-server-static-core` dependency version. 

 `i18next-http-middleware@3.7.1` requires `v5` , meanwhile  `i18next-http-middleware@3.6.0` is `v4`.

forgetso: Yes, this is fine.

packages/procaptcha-wrapper/vite.config.ts
forgetso: Our frontend vite config is full of all sorts of custom stuff for the bundle so a new one here is potentially ok. Is it a circular dep tho?

light-source: No, it isn't a circular one. The integration build config doesn't have any dependencies at all.

light-source: (I mean to the procaptcha packages. From the external it has vite for sure)

PR #1766, opened 2025-03-31
Drop unused translations

forgetso:


packages/util/src/util.ts
```diff
@@ -32,6 +32,46 @@ export const flatten = (obj: object, prefix = ""): Record<string, string> => {
 	return flattenedObj;
 };
 
+export const unflatten = (obj: Record<string, string>, prefix = "") => {
```
HughParry: Can we unit test unflatten please

packages/util/src/util.ts
```diff
@@ -32,6 +32,46 @@ export const flatten = (obj: object, prefix = ""): Record<string, string> => {
...
+	// biome-ignore lint/suspicious/noExplicitAny: <We dont know>
```
HughParry: Could we use generics to get around this or is it too much of a ballache?

packages/util/src/util.ts
```diff
@@ -32,6 +32,46 @@ export const flatten = (obj: object, prefix = ""): Record<string, string> => {
...
+	const unflattenedObj: Record<string, any> = {};
+
+	// Iterate through the entries of the object
```
HughParry: Lots of ott comments here haha

```
// Iterate through the entries of the object
for entry of object.entries

```

dev/config/src/vite/vite-plugin-remove-unused-translations.ts
```diff
@@ -0,0 +1,83 @@
...
+
+					// Remove keys that are not in `used`
+					const filteredData = Object.fromEntries(
+						Object.entries(jsonDataFlattened).filter(([key]) => used.has(key)),
+					);
+
+					const unflattened = unflatten(filteredData);
+
+					// Write back the filtered JSON
+					await fs.writeFile(
```
HughParry: We can speed up the build time by changing this to a promise.all - if it's building in ms already then nothing to bother with here tho

HughParry (changes requested):

packages/util/src/util.ts
forgetso: Completely changed function

PR #1861, opened 2025-05-21
fix/remove dev services

goastler:
- **remove dev service from docker compose**
- **add provider override compose file**


goastler:
needs testing on staging

PR #1862, opened 2025-05-21
reconfigure docker ports to stop providers from being accessed from the outside

goastler:


goastler:
needs testing on staging

forgetso:
Providers are not accessible from outside

PR #1863, opened 2025-05-23
Ensure imageThreshold is saved

forgetso:


forgetso:
Cypress tests pass locally.

PR #1911, opened 2025-06-05
feat/logging

goastler:
- **make logging package**
- **remove logging from common pkg**
- **npmignore**
- **logging using pino**
- **pkg lock**
- **adjust all logging call sites to structured logging**


packages/logging/src/logger.ts
```diff
@@ -0,0 +1,197 @@
...
+}
+
+const JSONReporter = (
+	logObject: LogObject,
+	context: {
+		options: ConsolaOptions;
+	},
+) => {
+	// https://stackoverflow.com/a/65886224
+	const writer = process?.stdout
```
forgetso: Check out main, this needs to be modified to check if `process === undefined`

forgetso (dismissed):

PR #1949, opened 2025-07-04
use redis non-stack

goastler:
use img `redis:8.0` for example


HughParry:
test it works on both staging and prod, then mark as ready for review

PR #2040, opened 2025-07-28
fix/lint imports

goastler:
- **remove circular dep on @prosopo/config package, no longer exists**
- **make lint package look at src code imports to detect missing and unneccessary deps**


goastler (commented):
.

goastler (commented):
.

goastler (commented):
.

PR #2041, opened 2025-07-28
feat/better logger

goastler:
- **make getLogger() not accept log level**
- **make getLogger accept a url and scope**
- **make getLogger strip file url prefix**
- **add subScope() to Logger**
- **use a clone method to set full logger config in NativeLogger**
- **add deepClone defaultData field of NativeLogger**
- **add event field to log records**
- **fix event handling in log method**
- **adjust getLogger() callsites to include scope and url**
- **remove auto-shorten url in logger - leave as is**
- **lint**


forgetso:
Re-open if needed

PR #2057, opened 2025-08-19
feat/access-policy-group

light-source:
https://github.com/prosopo/captcha-private/issues/2591


packages/common/src/logger.ts
```diff
@@ -347,7 +347,7 @@ export class NativeLogger implements Logger {
 	}
 }
 
-const stringifyBigInts = (value: unknown): unknown => {
+export const stringifyBigInts = (value: unknown): unknown => {
```
HughParry, line 359: Could you add jsdoc to this briefly explaining how/why casting to string please. Lack of docstring + types being unknown makes it fairly easy to accidentally misuse this func

packages/user-access-policy/src/api/deleteRuleGroupsEndpoint.ts
```diff
@@ -0,0 +1,101 @@
...
+
+	async processRequest(
+		args: DeleteRuleGroupsInputEndpointSchema,
+	): Promise<ApiEndpointResponse> {
+		const ruleIdPromises = [];
+
+		for (const ruleToDelete of args) {
+			const currentRuleIdPromises = ruleToDelete.clientIds.map(
+				async (clientId) => {
+					return await this.accessRulesStorage.findRuleIds({
```
HughParry: I don't think you have to do this awaiting inside the loop. Would be better to do something like:

```
const foundRuleIdsPromises = args.flatMap((ruleToDelete) =>
                                                                       ruleToDelete.clientIds.map((clientId) => 
                                                                       this.accessRulesStorage(etc etc)))

const foundRuleIds = Promise.all(foundRuleIdsPromises)
```

In fact with the awaiting inside the loop I'm not sure that the Promise.all like it's currently working is actually executing this fetch in parallel, but i could be misreading

packages/user-access-policy/src/api/deleteRuleGroupsEndpoint.ts
```diff
@@ -0,0 +1,101 @@
...
+			ruleIdPromises.push(...currentRuleIdPromises);
+		}
+
+		const foundRuleIds = await Promise.all(ruleIdPromises);
+		const ruleIds = foundRuleIds.flat();
+
+		// Set() automatically removes duplicates
+		const uniqueRuleIds = [...new Set(ruleIds)];
+
+		if (uniqueRuleIds.length > 0) {
```
HughParry, line 72: Could this be more like:

```
if (uniqueRuleIds.length < 0) {

		return {
			status: ApiEndpointResponseStatus.FAILURE,
			data: {
				deleted_count: 0,
                msg: "no rules to delete found :("
			},
		};
```

I'm just thinking that if no rules were found to delete, I'd want to know about it without having to check provider logs. Also happy if you throw an error or something

HughParry (dismissed):

packages/common/src/logger.ts
light-source: It makes sense, added

packages/user-access-policy/src/api/deleteRuleGroupsEndpoint.ts
light-source: `await` is indeed redundant in this case, but doesn't affect the concurrency. 

I've applied your suggestion with `.flatMap` for the cleaner code.

packages/user-access-policy/src/api/deleteRuleGroupsEndpoint.ts
light-source: Let me disagree on it. Imho, the 'failure' status should only be used in cases when the action has indeed failed, like Redis being unresponsive, or similar, in order to attract attention.

The case with no rules to delete doesn't seem to be a failure: it can be a new provider instance without the mentioned rule groups being present, and the command "get rid of it" in general should be claimed as "done", as after this request, the provider does have the desired state.    

The endpoint always returns the `deleted_count` in its response, so you can see how many, if at all, records were removed without checking any logs. It can also be handled conditionally, so if in the CLI for some reason we need attention on "empty cases", we can employ this variable. 

```
return {
			status: ApiEndpointResponseStatus.SUCCESS,
			data: {
				deleted_count: uniqueRuleIds.length,
			},
		};
```

Let me know if you don't share the point, and I'll change it.

HughParry: Yeah I think you're correct here. The cli that calls this should do some sort of 

```
if { !(res.data.deleted_count > 0) } 
   {throw new Error("nothing got deleted aaaaaaa")}
```

Which is something that can get sorted in the future

PR #2072, opened 2025-08-28
Secure random provider fixes

forgetso:
- 📦🔒
- Be more lenient with random provider selection

### DIFF
--- a/packages/database/src/databases/provider.ts
+++ b/packages/database/src/databases/provider.ts
@@ -1670,17 +1670,20 @@ export class ProviderDatabase
 		});
 	}
 
-	/** @description Remove a detector key */
+	/**
+	 * @description Remove a detector key
+	 * @param detectorKey The detector key to remove
+	 * @param expirationInSeconds Optional expiration time in seconds (default is 10 minutes)
+	 * */
 	async removeDetectorKey(
 		detectorKey: string,
 		expirationInSeconds?: number,
 	): Promise<void> {
 		const filter: Pick<DetectorSchema, "detectorKey"> = { detectorKey };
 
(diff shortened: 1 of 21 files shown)


packages/database/src/databases/provider.ts
```diff
@@ -1671,11 +1671,16 @@ export class ProviderDatabase
...
 
 	/** @description Remove a detector key */
-	async removeDetectorKey(detectorKey: string): Promise<void> {
+	async removeDetectorKey(
+		detectorKey: string,
+		expirationInSeconds?: number,
+	): Promise<void> {
 		const filter: Pick<DetectorSchema, "detectorKey"> = { detectorKey };
 
 		// Instead of deleting, set the expiresAt field to expire in 10 minutes
```
HughParry: Make comments more clear here

HughParry: In fact this should just be a nice docstring tbh

packages/provider/src/tasks/frictionless/frictionlessTasks.ts
```diff
@@ -131,6 +131,27 @@ export class FrictionlessManager extends CaptchaManager {
...
 
+	async scoreIncreaseUnverifiedHost(
+		host: string,
+		baseBotScore: number,
+		botScore: number,
+		tokenId: FrictionlessTokenId,
+	) {
+		this.logger.info(() => ({
+			msg: "Host not verified",
+			data: { requested: this.config.host, selected: host },
```
HughParry, line 142: Would it be possible to log the entropy here so we can better see what has gone wrong (cache issue vs. malicious actor is something i imagine i'll want to look into in the future)

HughParry (dismissed):

PR #2081, opened 2025-09-04
feat/pluggable-redis

light-source:
https://github.com/prosopo/captcha-private/issues/2642


packages/api/package.json
```diff
@@ -34,9 +34,8 @@
 	},
 	"homepage": "https://github.com/prosopo/captcha#readme",
 	"dependencies": {
-		"@prosopo/types": "3.1.4",
-		"@prosopo/user-access-policy": "3.5.10",
-		"@prosopo/config": "3.1.11"
+		"@prosopo/config": "3.1.11",
```
forgetso: Why have the deps been reduced?

packages/provider/src/tests/integration/api/blacklistRequestInspector.integration.test.ts
```diff
@@ -351,29 +360,7 @@ describe("blacklistRequestInspector Integration Tests", () => {
 			);
 			expect(nonClientResult.length).toBe(0);
 		});
-		it("should not return a match for a different IP", async () => {
```
forgetso, line 354: Why has this test been removed?

packages/provider/src/api/public.ts
```diff
@@ -35,9 +36,35 @@ export function publicRouter(): Router {
 	 */
 	router.get(PublicApiPaths.GetProviderDetails, async (req, res, next) => {
 		try {
-			return res.json({ version, ...{ message: "Provider online" } });
+			const db = env.getDb();
+
+			const redisConnection = db.getRedisConnection();
+			const redisAccessRulesConnection = db.getRedisAccessRulesConnection();
+
+			const response: ProviderDetails = {
```
forgetso, line 44: Is this introducing asynchronous operations into the details endpoint?

forgetso (changes requested):

packages/api/package.json
light-source: Only the UAP dependency was removed as no longer used. We discussed in Slack recently the move of the UAP-related methods to the UAP package

packages/provider/src/api/public.ts
light-source: No. Both are objects of the new connection type, and the endpoint uses their sync `isReady()` and `getAwaitingTimeMs() `methods :

```
export type RedisConnection = {
	isReady: () => boolean;
	getClient: () => Promise<RedisClientType>;
	getAwaitingTimeMs: () => number;
}; 
```

packages/provider/src/tests/integration/api/blacklistRequestInspector.integration.test.ts
light-source: it [was](https://github.com/prosopo/captcha/blob/main/packages/provider/src/tests/integration/api/blacklistRequestInspector.integration.test.ts) literally a duplication of the test above.

PR #2104, opened 2025-09-11
check access rules for provided ips

forgetso:
- **Check access rules for provided ips**
- **docs(changeset): Check access policies for provided IPs**

### DIFF
--- a/packages/provider/src/tasks/captchaManager.ts
+++ b/packages/provider/src/tasks/captchaManager.ts
@@ -30,10 +30,11 @@ import type {
 	UserScopeApiInput,
 	UserScopeApiOutput,
 } from "@prosopo/user-access-policy";
-import { getIPAddress } from "@prosopo/util";
-import { getPrioritisedAccessRule } from "../api/blacklistRequestInspector.js";
-import { getIpAddressFromComposite } from "../compositeIpAddress.js";
-import { deepValidateIpAddress } from "../util.js";
+import { flatten } from "@prosopo/util";
+import {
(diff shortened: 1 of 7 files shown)


packages/provider/src/tasks/captchaManager.ts
```diff
@@ -271,6 +272,24 @@ export class CaptchaManager {
...
 	}
 
+	async computeUserScopeAndGetAccessPolicy(
+		userAccessRulesStorage: AccessRulesStorage,
+		dapp: string,
+		headers: Record<string, unknown>,
+		ip?: string,
+		ja4?: string,
+		user?: string,
+	) {
```
HughParry, line 282: The AI comment below sounds reasonable. In addition i'd explicitly type the return of this func as a sanity check

packages/provider/src/tasks/powCaptcha/powTasks.ts
```diff
@@ -248,6 +252,16 @@ export class PowCaptchaManager extends CaptchaManager {
 				providedIp: getCompositeIpAddress(ip),
 			});
 
+			if (accessPolicy && accessPolicy.type === AccessPolicyType.Block) {
+				this.logger.info(() => ({
+					msg: "Provided IP is blocked by access policy",
+					data: {
+						providedIp: ip,
```
HughParry, line 259: log the access policy here for ez debugging

HughParry (changes requested):

forgetso:
Can revisit this in future if necessary

PR #2133, opened 2025-09-30
Fix/persist sessions for user ip combinations

HughParry:


HughParry:
Resolves https://github.com/prosopo/captcha-private/issues/2718

packages/provider/src/api/captcha/getFrictionlessCaptchaChallenge.ts
```diff
@@ -67,22 +69,20 @@ export default (
...
+					scoreComponents: {
+						baseScore: 0,
+					},
+					providerSelectEntropy: 0,
+					ipAddress: getCompositeIpAddress(req.ip || ""),
+					webView: false,
+					iFrame: false,
+					decryptedHeadHash: "",
+				});
+				return res.json(await tasks.frictionlessManager.sendPowCaptcha());
```
forgetso, line 85: is this meant to be here?

forgetso: Surely it would stop any further processing.

forgetso (changes requested):

PR #2210, opened 2025-11-05
make getDependencies handle non-zero exit code from npm commands

goastler:


goastler:
closing because we should have a valid pkg lock file rather than allowing the npm cmd to fail when invalid pkgs are involved

PR #2224, opened 2025-11-07
fix/entropy calcs pro

forgetso:
- **Calculate entropy for pro+ only**
- **docs(changeset): Run job for pro+ only**


HughParry:
@forgetso please don't merge. Would be useful to discuss this prior to merge

forgetso:
@HughParry  this is live, we need to get it merged, or some variation.

HughParry:
@forgetso merging this then

PR #2274, opened 2025-11-27
Feat/uaps with headhash and coords

HughParry:
- Added headHash and coords as queryable fields in user access policies
- Implemented user access policy checking in verifyPowCaptchaSolution to block users with hard-block policies (ignores policies with captchaType set)


packages/user-access-policy/src/redis/reader/redisRulesQuery.ts
```diff
@@ -135,9 +166,10 @@ const getPolicyScopeQuery = (
 	const clientId = policyScope?.clientId;
 
 	if ("string" === typeof clientId) {
+		const escapedClientId = escapeTagValue(clientId);
 		return FilterScopeMatch.Exact === scopeMatch
-			? `@clientId:{${clientId}}`
-			: `( @clientId:{${clientId}} | ismissing(@clientId) )`;
+			? `@clientId:{${escapedClientId}}`
```
forgetso: Why is clientId needing escaped all of a sudden?

packages/provider/src/tasks/powCaptcha/powTasks.ts
```diff
@@ -53,6 +75,53 @@ export class PowCaptchaManager extends CaptchaManager {
...
+		let headHash: string | undefined;
+		if (challengeRecord.sessionId) {
+			const sessionRecord = await this.db.getSessionRecordBySessionId(
+				challengeRecord.sessionId,
+			);
+			headHash = sessionRecord?.decryptedHeadHash;
+		}
+
+		// Serialize coords to string for querying
+		const coordsString = coords ? JSON.stringify(coords) : undefined;
```
forgetso, line 101: For now this is ok but we need a follow up which is looking for approximate coords. So coords with a threshold similarity. This is because a site might have the widget located at [120,80] for everyone and an attacker have hosted the widget in a completely different position in a faked origin site. Mouse jitters would overcome this check even though they're still blatantly in the wrong area overall.

packages/provider/src/tasks/powCaptcha/powTasks.ts
```diff
@@ -156,6 +228,43 @@ export class PowCaptchaManager extends CaptchaManager {
 			}
 		}
 
+		// Check user access policies for hard blocks
+		if (userAccessRulesStorage) {
```
forgetso: Needs moved to verify as per comment in Slack

packages/provider/src/tasks/powCaptcha/powTasks.ts
HughParry: True, though I worry that we'll end up accidentally blocking everybody who's rendering the site differently to most if we go hard here (e.g. massive tablet rendering a giant mobile view vs desktop vs small mobile etc)

forgetso: This kind of logic would be reserved for cases where we were absolutely certain of the outlier - e.g. couple it with attributes that have only been seen for a clump of traffic that has just recently switched on.

forgetso (dismissed):

PR #2282, opened 2025-12-04
add unit tests to util pkg

goastler:


packages/util/src/fetchWithEtag.unit.test.ts
```diff
@@ -0,0 +1,214 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+import { fetchWithETag } from "./fetchWithEtag.js";
+
+describe("fetchWithETag", () => {
```
HughParry: This unit tests demonstrates that the util is a bit mental imo. Will remove and change the code where it is used

packages/util/src/solverService.unit.test.ts
```diff
@@ -0,0 +1,144 @@
...
+
+	test("returns different nonces for different data", () => {
+		const data1 = "test1";
+		const data2 = "test2";
+		const difficulty = 2;
+
+		const nonce1 = solvePoW(data1, difficulty);
+		const nonce2 = solvePoW(data2, difficulty);
+
+		// They might be the same by chance, but likely different
```
HughParry: Remove this test or replace it with a valid one where the nonces found are verifiably different. Otherwise, this is no different to running the solvepow for difficulty 2 test

HughParry (changes requested):

packages/util/src/fetchWithEtag.unit.test.ts
HughParry: Resolved by https://github.com/prosopo/captcha-private/pull/3015, ready for review

packages/util/src/solverService.unit.test.ts
goastler: addressed, pls review

PR #2291, opened 2025-12-04
test account pkg

goastler:


HughParry (changes requested):
This fails since no changeset and no vite.test config file has been added

PR #2362, opened 2026-01-20
Feat/decision machine

HughParry:


packages/database/src/databases/provider.ts
```diff
@@ -1816,6 +1825,51 @@ export class ProviderDatabase
...
+				},
+				$setOnInsert: {
+					createdAt: now,
+				},
+			},
+			{ upsert: true },
+		);
+	}
+
+	async getDecisionMachineArtifact(
```
forgetso, line 1875: Its not entirely clear that a single decision machine is fetched whenever this functionality kicks in. I can imagine us easily having decision machines that were specific to device type, even if its just mobile/app/desktop initially. For now, I think there are some naming changes that could be made to indicate that there is one global decision machine.

forgetso (approved):

forgetso (changes requested):
see comment regarding scope

forgetso (approved):

PR #2579, opened 2026-05-22
feat(health): add consistent /health and /auth/health endpoints to provider API

goastler:
## Summary

- Adds `HealthApiPaths` enum, `HealthResponse`, and `AuthHealthResponse` types to `@prosopo/types`
- Implements `GET /health` (public, returns `{ alive: true, timestamp }`) on the Provider API
- Implements `GET /auth/health` (JWT auth required, returns health + version, uptime, service name) on the Provider API

Related to prosopo/captcha-private#1469


forgetso (commented):
## Review: `/health` & `/auth/health` on Provider API

Companion to prosopo/captcha-private#3372 — leaving the bulk of the cross-service review there.

### Provider-specific concerns

1. **Overlap with `/healthz`.** After this merges the provider exposes both `PublicApiPaths.Healthz` (`/healthz` → plain `OK`) and `HealthApiPaths.Health` (`/health` → JSON). These do the same job. Recommend dropping `PublicApiPaths.Healthz` and migrating these call sites:
   - `docker/docker-compose.provider.yml:28,63` — Docker healthchecks for provider1/provider2 (curl `/healthz`)
   - `demos/cypress-shared/scripts/wait-for-services.ts:23`
   - `architecture.mmd:34` (the `Healthz[\"GET /healthz\"]` node)
   - In captcha-private: `deploy-production.sh:48`, `.github/workflows/deploy-procaptcha-bundle.yml:179-182`, `packages/infrastructure/ansible/playbooks/provider.yml:196`, `packages/e2e/src/utils/start-services.ts:55`, `packages/e2e/src/utils/restart-services.ts:47`, `packages/e2e/E2E_TESTING_GUIDE.md`
   
   The new JSON response would need a string compare in the docker/curl healthchecks (e.g. `curl --fail localhost:9229/health | grep -q '\"alive\":true'`), or just rely on the 200 status.

2. **`apiApp.use(\"/auth\", authMiddleware(...))`** at `startProviderApi.ts` reserves the entire `/auth/*` prefix for JWT-authenticated routes. That's a sensible convention but worth calling out explicitly so future contributors don't accidentally mount a public route there. Consider a comment, or splitting into `/auth/health` mount instead of a path-prefix `use`.

3. **Type-only import ordering** — `import { authHealthRouter } from \"./authHealth.js\";` in `startProviderApi.ts` is inserted out of alphabetical order relative to neighbouring imports. Biome will likely flag it.

4. **`env` parameter is unused** in `authHealthRouter(env: ProviderEnvironment)` — drop the parameter unless something concrete is planned (e.g. surfacing DB/Redis readiness in `AuthHealthResponse`, which would actually be more useful than `process.uptime()` for monitoring).

5. **No tests** for the new endpoints — a small fixture asserting `200` for `/health` and `200/401` for `/auth/health` would be valuable.

6. **Auth scheme consistency note** — in the captcha-private PR, `monitoring`'s `/auth/health` uses a static bearer token (`HEALTH_API_TOKEN`) instead of JWT. Worth aligning all three services on the same auth method (JWT recommended) so a single probe token works platform-wide.

PR #2673, opened 2026-06-11
refactor: decouple error classes from i18n + conventional locale structure

goastler:
## Summary

Decouples the `Prosopo*Error` classes from i18n and logging, moves translations into a conventional i18n structure, and normalizes the error-class API so every error has a queryable translation key.

### Error model
- `ProsopoBaseError` and subclasses no longer translate or log at construction time. They carry a `translationKey` plus a fallback `message`; translation happens at the presentation layer (UI render or HTTP response via `unwrapError`).
- **Uniform constructor**: every error takes a required `TranslationKey` first arg and an optional causing `Error` via `options.cause` (whose message becomes the fallback). Removed the old `Error | TranslationKey` first arg, `options.translationKey`, and the `i18n`/`logger`/`logLevel` options.
- **`ProsopoApiError` accepts `TranslationKey` only** (not `Error`) — passing a raw `Error` leaked internal text into HTTP responses and produced no client key. `ValidErrorKey` / `BACKEND_ERROR_KEYS_ARRAY` remain the curated set preserved in the frontend bundle by the Vite plugin.
- Internal/CLI errors with no user-facing key use a `GENERAL.UNKNOWN` placeholder with the detail in `options.message` (e.g. env "not setup" guards). datasets-fs sites kept their existing `FS.*`/`DATASET.*` keys.

### Type-safe translation keys
- `disapproveDappUserCommitment(reason?)` is typed `ResultReason` (matching the persisted `result.reason`).
- Added the backend translation keys referenced in code but missing from the locale files, and propagated the full `en` key set to all 32 locales so the locale-consistency test passes.

### Tests
- Rewrote `@prosopo/common` error tests for the decoupled/uniform API.
- Updated `@prosopo/api-express-router` and provider integration tests for the new error shape (responses carry the locale-stable key, not translated text).

## Test plan
- `build:tsc` green across all buildable backend/non-frontend packages.
- `@prosopo/common`, `@prosopo/locale`, `@prosopo/database`, `@prosopo/api-express-router`, `@prosopo/logger`, `@prosopo/env`, `@prosopo/datasets-fs` unit tests pass; full `@prosopo/provider` unit suite (785) passes.
- Frontend widget packages and demos validated by CI.


goastler:
Thanks for the review — addressed all comments in 46b2538:

- **`errorKeys.ts` — `_validateErrorKeysExistInTranslations` was a no-op cast.** Replaced with a real compile-time check: `_validateErrorKeysExistInTranslations: readonly TranslationKey[] = BACKEND_ERROR_KEYS_ARRAY`, which fails to compile if any `ValidErrorKey` is missing from the locale JSON.
- **`loadI18next.ts` — concurrent callers could get a half-ready instance.** Both loaders now only short-circuit on `instance?.isInitialized`; while init is in flight they fall through to the shared promise and await readiness.
- **`loadI18next.ts` — rejected promise was never cleared.** Added a `.catch` that resets the cached promise/flag so initialization can be retried after a transient failure.
- **`errorHandler.ts` — `request.logger` could be absent.** Guarded with `request.logger?.error(...)` so a missing request-logger middleware no longer throws and masks the original error.
- **`validateAddress.ts` — unused `logger` param.** Renamed to `_logger` (kept for call-site compatibility; errors no longer log at construction).
- **PR description mismatch.** Updated the description to reflect the final API (`ProsopoApiError` takes `TranslationKey` only).

PR #2798, opened 2026-07-06
feat: improved captcha look

HughParry:
CI Check Do Not Merge


HughParry:
https://github.com/user-attachments/assets/5bf87d58-19fe-42d7-9458-ad5c811442fe

HughParry:
## What changed

Merged `main` and resolved the conflict. Below is the split you asked for — styling vs structural — with the honest answer that there **are** a few structural changes, all small, plus one I'd flag for a decision.

### Markdown / docs: none 🎉

The only `.md` file touched is `.changeset/improved-captcha-look.md`, which is the changeset for this PR. No docs, READMEs or other markdown were restructured.

---

## 1. Styling changes (the bulk of the PR)

A Material 3 purple tonal theme. Values only — no API shape involved.

**`widget-skeleton/src/theme.ts`** — the substance of the PR:
- Purple tonal palette replacing the previous blue/grey (`primary.main` `#487DFA` → purple ramp)
- M3 surface roles: `surface`, `surfaceContainerHigh`, `onSurface`, `onSurfaceVariant`, `primaryContainer`
- State-layer opacities (`stateLayer.hover/focus/pressed`) instead of shadow-based elevation
- An M3 type scale slice (`labelLarge`, `bodyMedium`, `titleMedium`)
- `error.main` `#f44336` → `#b3261e` (M3 error role)
- Elevation tokens set to `none`, kept as tokens so a shadowed scheme can be reinstated in one place

**`widget-skeleton/src/constants.ts`** — `WIDGET_BORDER_RADIUS` `8px` → `16px`.

**`widget-skeleton/src/elements/`** — `skeleton.ts` moves the shell onto `surface`/`border` and adds a shadowless hover state layer; `checkbox.ts` redraws the spinner as an M3 circular progress (neutral track, primary active arc).

**Components** — `Button.tsx`, `Checkbox.tsx`, `Reload.tsx`, `CaptchaComponent.tsx`, `CaptchaWidget.tsx`, `Modal.tsx`, `PuzzleCanvas.tsx` consume the new tokens. `Button.tsx` also gains an M3 focus indicator (3dp outline, 2dp offset) driven by `:focus-visible`.

---

## 2. Structural changes (four, each with a reason)

**a. `withAlpha(hex, opacity)` — new exported helper in `theme.ts`**
Five lines, pure. Needed because M3 state layers and translucent tokens are colour-plus-opacity, and the alternative is hardcoded `rgba()` literals that no longer track the palette. Picked up automatically by the existing `export * from "./theme.js"`, so `index.ts` is untouched.

**b. `PuzzleCanvas` gains a required `theme: Theme` prop** (and `ProcaptchaWidget` passes it — the single `+1` line)
Unavoidable: the canvas previously hardcoded its colours, and a component can't be themed without receiving the theme. Importing `lightTheme` directly instead would have broken dark mode.

**c. New palette keys** — `surface`, `onSurface`, `onSurfaceVariant`, `surfaceContainerHigh`, `primaryContainer`, `puzzle`, plus `stateLayer` and `typography`
Additive. `Theme` is inferred (`typeof lightTheme | typeof darkTheme`) and nothing outside `widget-skeleton` constructs a theme literal, so this widens the type without breaking consumers.

**d. ⚠️ `palette.grey` was removed** — worth a decision
The grey scale was on the palette on `main` and is gone here. Nothing in this repo reads `palette.grey`, so it's safe internally — but `@prosopo/widget-skeleton` is **published** (`2.8.5`, not private), so an external consumer reading `theme.palette.grey[300]` would break. It's the only genuinely breaking removal in the PR. Happy to restore it as a deprecated passthrough if you'd rather keep the surface stable; I left it alone since dropping it looks deliberate.

---

## 3. Conflict resolution

One conflict, in `procaptcha-react/src/components/Button.tsx`, and only in the import block. `main` added `isEventTrusted` (used by the click guard) while this branch added `withAlpha`. Both were kept — `main`'s `if (!isEventTrusted(e))` guard is preserved verbatim.

## 4. Test updates

`#2930` landed tests on `main` that pin *which palette key* each element interpolates. Three named keys this branch deliberately moved away from:

- The widget shell now reads `surface`, not `background.default`. Both still exist — under M3 `background.default` is the **dialog** container's role (`surfaceContainerHigh`), which is what separates a resting affordance from a modal container.
- The checkbox spinner is now an M3 circular progress (`border` track + `primary` arc) where it previously drew the whole ring in `background.contrastText`.

The assertions were pointed at the new tokens. **No styling was changed to satisfy a test** — in each case the element moved first and the test followed.

## 5. Verification

`turbo run test` **83/83 tasks**, `turbo run typecheck` **49/49**, `npm run lint` clean (biome + licence + refs + testCheck). Also ran the stack locally and eyeballed the widget across the demo pages.

https://claude.ai/code/session_017hPGTJ23KyDz2WuUgzz2pi

PR #2812, opened 2026-07-08
feat(lb-weights): metric-driven Bunny DNS load-balancer weight updater

goastler:
## Summary

Adds a new package **`@prosopo/lb-weights`** — a service that dynamically sets Bunny DNS load-balancer record weights from per-node Prometheus metrics.

On a fixed interval it:
1. reads Bunny DNS zones and filters to the load-balancer address pools for the configured subdomains;
2. scrapes each node's `<ip>/metrics` Prometheus endpoint;
3. inverse-weights a chosen metric (lower usage → higher weight);
4. applies integer weights to Bunny — unless in dry-run (readonly) mode.

Defaults are safe: **dry-run is on by default**, interval 15s, metric `cpu_usage_1m`.

## Weighting

- Inverse scheme: negate → shift by min → `+1` (never 0) → scale proportionally into Bunny's max → `ceil` + clamp.
- **Bunny constraints handled**: weight is an integer in `[0, 100]` (confirmed from Bunny docs); computed weights are scaled to preserve ratios and clamped to `[1, 100]`.
- **Missing-metric fallbacks**: an unreachable node defaults to the pool's **max weight**; if every node in a pool is unreachable, **equal weighting** is used.
- **Spike detection**: logs a warning for any weight `> mean + 1σ`, and a distinct louder warning for `> mean + 2σ` (max-side only, per pool).

## Observability

- Structured logging via `@prosopo/logger` (which pool is computed/updated, per-node ip/metric/weight, dry-run, mode).
- `docker-compose.yml` in the package dir wires the service + **Vector** + **OpenObserve**, JSON logging on all services, `env_file` support, and basic-auth sink credentials (`vector.toml`).
- `Dockerfile` and `.env.template` included in the package.

## Tests

52 unit + mocked-integration/e2e tests, all passing. **No test hits the real Bunny API or real `/metrics` endpoints** — all external calls are injected/mocked.

## Notes

- Package built with `build:tsc` (type-checked); tests run under vitest.
- New workspace dependency: `@prosopo/logger`.

### DIFF
--- a/packages/lb-weights/.env.template
+++ b/packages/lb-weights/.env.template
@@ -1,0 +1,40 @@
+# Copy to `.env` and fill in. Used by docker-compose.yml and the lb-weights service.
+
+# ── Bunny DNS ───────────────────────────────────────────────────────────────
+# Bunny DNS API token. Needs read access to zones and (unless dry-run) write
+# access to set record weights.
+BUNNY_API_KEY=
+
+# ── Load balancer targets ────────────────────────────────────────────────────
+# Comma-separated subdomains (record names) whose load-balancer pools to manage.
+LB_SUBDOMAINS=pronode
+
+# ── Weighting ────────────────────────────────────────────────────────────────
+# Prometheus metric that drives weighting. Lower value => higher weight.
+LB_METRIC_NAME=cpu_usage_1m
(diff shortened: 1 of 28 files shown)


goastler:
Wrong repo — this package belongs in prosopo/captcha-private, not prosopo/captcha. Reopening there.

PR #2907, opened 2026-07-27
chore(deps): bump vite to 6.4.3 and mongoose to 8.24.1

HughParry:
Companion to prosopo/captcha-private#3916, which batches ~21 Dependabot PRs into one lockfile regeneration. Two of those bumps cannot land in captcha-private alone — they need this submodule to move in lockstep.

## Why this repo has to move too

`captcha-private` treats `captcha/packages/*` as npm workspaces, so a version pinned here competes with the parent's pin for the hoisted slot:

- **vite** — this repo pinned `6.4.1` in 44 manifests. When the parent bumped to `6.4.3`, npm hoisted one and nested the other, so `tsc` saw two structurally distinct `Plugin<any>` types and `captcha/dev/config` failed with `TS2345`/`TS2769`. That is what made Dependabot's own vite PR (captcha-private#3900) red.
- **mongoose** — same story: this repo pinned `8.13.0` in 6 manifests, giving a hoisted 8.13.0 alongside a nested 8.24.1 *and* two `mongodb` copies (6.15.0 / 6.20.0). `BulkWriteResult` then resolved to two different nominal types and `database-private` failed to compile.

Bumping both here de-duplicates them. The parent lockfile actually **shrinks** as a result (66,604 → 63,485 lines) rather than growing by ~23k, which is what happened when only the parent moved.

## Type changes

mongoose 8.24 changed `Document`'s id parameter from `any` to `Types.ObjectId`. That surfaced code that only ever compiled because the old default was `any`:

- `import { ObjectId } from "mongoose"` is the **schema** type (`Schema.Types.ObjectId`), not the id `Document._id` carries. Switched to `Types.ObjectId` in `types-database/src/types/provider.ts` and `database/src/databases/provider.ts`.
- `provider/src/util.ts` cast `runningTask._id as ObjectId`; `_id` is already the right type, so the cast is gone.
- Two mocks in `clientTasks.unit.test.ts` used `_id: 0`; now `new Types.ObjectId()`. The assertions only check `save` was called, so this is inert.

## Verification

Run from the parent repo with this submodule checked out at this branch:

- `turbo run typecheck` — 87/87
- `turbo run build:tsc` — 79/79
- `biome check .` — clean

One `vite@6.4.1` copy remains, nested under `packages/fingerprintjs` (a separate submodule, left untouched deliberately to avoid dragging a third repo into this). It is isolated and does not affect any build.

https://claude.ai/code/session_01GagGdoZ8Lyj876K5EsDwRb


HughParry:
Superseded by #2850, which carried this branch's work: the mongoose 8.24.1 / mongodb 6.20.0 bumps and all four Types.ObjectId / cast / mock fixes are already on main and byte-identical there.

Nothing is left to salvage — every remaining conflict is this branch's older pins losing to main's (vite 6.4.3 vs 8.1.5, vitest 3.2.x vs 4.1.10, @vitest/coverage-v8, @vitejs/plugin-react, @vitejs/plugin-vue), plus a revert of the workspaces change that drops integration/frameworks. Resolving them all in main's favour would leave an empty PR.

PR #2942, opened 2026-07-31
fix(provider): add decodePayload.d.ts to stop TS inferring from obfuscated JS

forgetso:
## Summary
- Adds a hand-authored `.d.ts` companion for `decodePayload.js` so TypeScript never has to infer types from the obfuscated file.
- Same pattern already used by `decodeBehavior.js`/`.d.ts` and `decodeSimd.js`/`.d.ts` — `decodePayload.js` was the odd one out.

## Why
`packages/catcher/vite.provider.config.ts` rewrites `packages/provider/src/tasks/detection/decodePayload.js` via `viteStaticCopy` on every build:provider run. `js-confuser` uses a random seed, so the output shape varies run-to-run. In the [v3.6.75 publish run](https://github.com/prosopo/captcha-private/actions/runs/30667988337/job/91279347477), the obfuscated JS defeated TS's default-export inference and the whole `aws-serverless-api` `build:tsc` failed:

```
getBotScore.ts(16,8): TS7034 Variable 'getBotScoreFromPayload' implicitly has type 'any'
getBotScore.ts(23,24): TS7005 Variable 'getBotScoreFromPayload' implicitly has an 'any' type
getBotScore.unit.test.ts:*: TS2571 Object is of type 'unknown'
```

Adding the `.d.ts` pins the exported type to the real source signature (`packages/catcher/src/integrity/node/getBotScore.ts`) regardless of what the obfuscator emits.

## Test plan
- [x] Ran `catcher bundle:provider` (regenerates the obfuscated JS with a fresh seed) three times, then rebuilt provider from scratch with `tsc --build --verbose` — all three iterations pass cleanly.
- [x] Full `aws-serverless-api build:tsc` from clean dist — passes.
- [x] `biome check` on the new file — clean.


forgetso:
Superseded — moving fix into captcha-private build process to avoid publishing detector-related types on the public repo.

PR #2961, opened 2026-08-02
test(types-database): unit and type tests for the mongoose schemas

goastler:
Wires `@prosopo/types-database` into CI (`test` script + `vite.test.config.ts`) and adds 91 tests — 78 unit, 13 type.

## Covered

- **Client settings defaults** — every default a client inherits when it sends nothing, pinned against the shared constants in `@prosopo/types` so a drift in either shows up. Includes the traffic filter (only `blockAbuser` is on by default), the 0–1 abuser score range, `skipExtrasOnValidDnsPath`, the honeypot encoding enum, and the context-aware map being pre-populated while disabled.
- **IP validation rules** — defaults, negative-value rejection, per-country overrides, and `countryOverrides` staying `undefined` rather than an empty map (the read path treats a present map as an intentional override set).
- **The composite IP Decimal128 setter** — the interesting logic in `provider.ts`. An IPv6 half arrives as a bigint, a BSON `Long`, or a string, and all three have to land on the same value: 2^64-1 stays exact rather than rounding through a double, a `Long` comes back unsigned rather than negative, and a `Long` from a *different* bson copy is still recognised by its `_bsontype` brand (an `instanceof` check would silently skip normalisation).
- **Index intent** — that the pending-stage sweep is a partial index, that the `$exists: false` backfill indexes are deliberately *not* partial, the detector TTL index, and the unique constraints.
- **TTLs** — user solutions and stored captchas expire after four weeks, scheduled tasks after a week, and an unstaged captcha never expires because `storedAtTimestamp` is unset.
- **Record validation** — required fields, enum rejection, and the label fields staying empty until a human sets them.
- **Type tests** — the `IDatabase` contract (`getConnection()` returns a non-optional `Connection` even though `connection` is optional), the domain record shapes, `StoredSession` still being the very same type as `SessionRecord` after the schema merge, and which `UserAgentInfo` fields are guaranteed.

## Noted, not fixed

`PoWCaptchaRecordSchema.result.reason` sets `enum: TranslationKeysSchema.options`, but that list comes back **empty**, so the constraint is inert and any string is accepted. The cause is `getLeafFieldPath` in `@prosopo/locale`, which returns `[]` for a leaf and therefore drops every key. Fixing it would narrow `TranslationKey` from `string` to a real union and break callers across the repo, so it belongs in its own change to that package. A test pins the current behaviour with this explanation rather than asserting an enforcement that does not happen.

### DIFF
--- a/packages/types-database/package.json
+++ b/packages/types-database/package.json
@@ -20,6 +20,7 @@
 		"clean": "del-cli --verbose dist tsconfig.tsbuildinfo",
 		"build": "npm run build:cross-env -- --mode ${NODE_ENV:-development}",
 		"build:cross-env": "vite build --config vite.esm.config.ts",
+		"test": "NODE_ENV=${NODE_ENV:-test}; npx vitest run --config ./vite.test.config.ts",
 		"build:tsc": "tsc --build --verbose",
 		"build:cjs": "NODE_ENV=${NODE_ENV:-development}; vite build --config vite.cjs.config.ts --mode $NODE_ENV",
 		"typecheck": "tsc --project tsconfig.types.json"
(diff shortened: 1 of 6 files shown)


goastler:
Superseded by https://github.com/prosopo/captcha/pull/2951, which already covers this package. The extra cases here will be reopened on top of it once it lands.

PR #3011, opened 2026-08-06
feat(provider): session chain — origin fallback for DM-input reads

forgetso:
## Summary

Escalation sessions inherit an `originSessionId` back to the pow-solve that spawned them. A new `CaptchaManager.getSessionRecordWithOriginFallback` walker fills in fields that are inherently populated on the origin — `simdReadings` (attached fire-and-forget on pow-submit) and `dnsEvent` (set by the DNS sidecar on the origin's TLS connection) — when the escalation record lacks them. The three `serverVerify*CaptchaSolution` methods use the walker so decision-machine inputs on escalated puzzle / image sessions see the origin's SIMD readings and DNS event instead of `undefined`.

## Observed impact (before fix)

Field-presence on `sessions.createdAt > now-60m`, representative pronode:

| field | non-escalation | escalation |
|---|---|---|
| `simdReadings` | 53 % | **3 %** |
| `dnsEvent` | 99 % | **2 %** |
| `bundleId` | 95 % | 98 % |
| `decryptedHeadHash` | 95 % | 100 % |

Split by escalation captchaType (last 60m, pronode10): `puzzle` (n=25) had 2 with simd / 1 with dns. `image` (n=32) had 0 / 0. This was tripping decide-machine deny rules (`SIMD_ABSENT` etc.) on legit escalation flows.

Root cause: `verifyPowCaptchaSolution` awaits Redis for the SIMD attach but fires the Mongo update through `scheduleMongoSimdReadingsUpdate` (fire-and-forget). `buildEscalation` then reads the origin from Mongo — the async write hasn't landed. `dnsEvent` is worse: the sidecar patches it onto the origin session directly and `buildEscalation`'s `createSession` call never forwarded it.

## Change

- Add `originSessionId: string` to `SessionRecordSchema` and the projection in `getSessionRecordBySessionId`.
- `buildEscalation` in `submitPoWCaptchaSolution.ts` sets `originSessionId: originSession.sessionId` on the new session.
- New `CaptchaManager.getSessionRecordWithOriginFallback(sessionId)`: reads the session, and if it's an escalation with `originSessionId` AND any fallback-eligible field is missing, walks to the origin and fills the gap. Escalation-owned fields (`captchaType`, `sessionId`, `score`, `ipInfo`, `headers`) are never overridden.
- `PowCaptchaManager.serverVerifyPowCaptchaSolution`, `PuzzleCaptchaManager.serverVerifyPuzzleCaptchaSolution`, `ImgCaptchaManager.serverVerifyImgCaptchaSolution` all switch their DM-input session read to the walker.

## Non-goals

- No change to write paths for `simdReadings` / `dnsEvent` — origin remains the single source of truth, escalation just references it.
- Non-escalations and legacy escalations without `originSessionId` are pass-through no-ops.
- Extra Mongo read only fires when the escalation is actually missing a fallback-eligible field.

## Test plan
- [x] `npm run -w @prosopo/provider test:unit` — `captchaManager.unit.test.ts` passes 55/55 (48 pre-existing + 7 new walker tests).
- [x] Type-check clean: `types`, `types-database`, `database`, `provider`.
- [ ] Post-deploy: `session.simdReadings` / `session.dnsEvent` presence rate on escalation sessions climbs from ~3 % / ~2 % to match the non-escalation baseline (~53 % / ~99 %). Decide-machine `SIMD_ABSENT`-style deny counts on escalated puzzle / image sessions falls to near-zero.

## Regression guards

- Integration: `sessionRecordProjection.integration.test.ts` — asserts `originSessionId` round-trips through the projection.
- Unit (7 cases in `captchaManager.unit.test.ts`):
  - non-escalation session: pass-through, no origin lookup
  - session not found: undefined
  - all fallback fields present on escalation: no origin lookup fires
  - missing `simdReadings`: filled from origin
  - origin also lacks the fields: no fill
  - origin session deleted: pass-through escalation unchanged
  - escalation-owned fields (`captchaType`, `sessionId`, `score`) preserved even when origin has different values

### DIFF
--- a/.changeset/session-chain-origin-fallback.md
+++ b/.changeset/session-chain-origin-fallback.md
@@ -1,0 +1,16 @@
+---
+"@prosopo/provider": patch
+"@prosopo/database": patch
+"@prosopo/types-database": patch
+"@prosopo/types": patch
+---
+
+feat(provider,database,types): session chain — escalations reference origin, DM-input reads walk back for missing fields
+
+Adds `originSessionId` to the Session schema and populates it on escalation sessions in `submitPoWCaptchaSolution.buildEscalation`. Adds `CaptchaManager.getSessionRecordWithOriginFallback` — a session reader that, when the record is an escalation missing an inherently-origin-populated field (`simdReadings`, `dnsEvent`, `entropyMathRandom*`, `entropyCrypto*`, `entropyWallClockOffsetMs`), reads the origin session and fills the gap. Escalation-owned fields (`captchaType`, `sessionId`, `score`, `ipInfo`, `headers`, etc.) are never overridden.
+
+The three `serverVerify*CaptchaSolution` methods now use the walker instead of the raw `getSessionRecordBySessionId`, so decision-machine inputs on escalated puzzle / image sessions see the origin's SIMD readings and DNS event.
+
+Fixes the write-time race between (a) the origin's fire-and-forget SIMD attach via `scheduleMongoSimdReadingsUpdate` on pow-submit, and (b) `buildEscalation`'s immediate Mongo read — which left ~97% of escalation sessions with no `dnsEvent` and ~97% with no `simdReadings`, in turn tripping decide-machine deny rules (SIMD_ABSENT etc.) on legit escalation flows.
+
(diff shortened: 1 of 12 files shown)


forgetso:
Folded into #3012 (squash-merge d6cb841e6). The session-chain-origin-fallback diff (commit f3ebed6b) is now on main via the #3012 squash.

PR #3031, opened 2026-08-11
fix(config): keep the server-only i18next backend out of frontend bundles

HughParry:
## Problem

The `i18nBackend` chunk in `procaptcha-bundle` is **153,593 bytes raw / 40,785 gzip** and contains nothing a browser can use:

- `i18next-fs-backend` — reads locale JSON off disk, and drags in its YAML (83,189 bytes), JSONC (52,686) and JSON5 (41,114) parsers
- `i18next-http-middleware` — Express request handling, with cookie/session/header language lookups

`i18next-fs-backend` has been on the frontend config's exclusion list for a long time. It didn't work, because **`external` only ever matches a specifier exactly**, and `i18nBackend.ts` imports the subpath:

```ts
import FSBackend from "i18next-fs-backend/cjs"; // i18next/i18next-fs-backend#57
```

`"i18next-fs-backend/cjs" !== "i18next-fs-backend"`, so it was bundled regardless. `i18next-http-middleware` was never on the list at all.

## Fix

1. `external` matches subpaths as well as bare names.
2. `i18next-http-middleware` joins the exclusion list.

I verified the blast radius rather than assuming it: I instrumented `external` to log every specifier where subpath matching would change the answer, and rebuilt the whole bundle. **`i18next-fs-backend/cjs` is the only one.** The filters are substring matches, so the two i18next entries catch only themselves — `i18next-http-backend`, `i18next-browser-languagedetector`, `i18next-chained-backend` and `i18next-resources-to-backend` are used in the browser and remain bundled (confirmed present in `i18nFrontend-*.js` after the change).

## Measured

`@prosopo/procaptcha-bundle`, `NODE_ENV=production`, gzip level 9.

| | before | after | delta |
| --- | ---: | ---: | ---: |
| `i18nBackend` chunk, raw | 153,593 | 788 | −152,805 (−99.5%) |
| all chunks, raw | 1,247,449 | 1,094,644 | −152,805 (−12.2%) |
| all chunks, gzip | 474,695 | 434,415 | −40,280 (−8.5%) |
| all chunks, brotli | 426,952 | 394,257 | −32,695 (−7.7%) |
| critical path, gzip | 199,625 | 199,639 | +14 (chunk-name hash length) |

**This is not a first-paint win.** The chunk is only reachable via `loadI18next(true)`, and the widget always calls `loadI18next(false, language)` — its only `backend: true` caller is `i18nMiddleware.ts`, which is server-side. This removes dead weight from the published artifact.

## Why this can't regress the browser

The chunk was already impossible to load in a browser. On `main` its only bare import is:

```
from "node:path"
```

`node:path` is already external, so any browser attempt to evaluate that chunk already failed. Adding two more externals to a chunk that never loads removes bytes without removing behaviour.

After the change the chunk is 788 bytes — the `initializeI18n` function plus three bare imports.

## Verification

- [x] `npm run -w @prosopo/config build:tsc`
- [x] `npm run -w @prosopo/config test` — 156 passed
- [x] `npx biome check` on changed files
- [x] Instrumented probe over the full module graph to confirm exactly one newly-excluded specifier
- [x] Confirmed browser-side i18next packages still bundled after the change


HughParry:
Pushed 49ff224a9 after CI caught a real regression in the first approach.

**What broke:** the original commit matched `pkg/*` against the whole exclusion list. The `vite` filter is a *substring* match, so it also selects `vite-plugin-node-polyfills` — and `vite-plugin-node-polyfills/shims/process` is injected into the served and IIFE bundles as a bare import. Externalising it left the browser unable to resolve the specifier, and all three cypress specs failed in `before each`:

```
Failed to resolve module specifier "vite-plugin-node-polyfills/shims/process".
Relative references must start with either "/", "./", or "../".
```

My pre-push probe missed it because it only instrumented the **ESM** build, where that specifier is resolved to an absolute path by the plugin before `external` is consulted. The served/IIFE path passes it through bare.

**Fix:** list `i18next-fs-backend/cjs` literally instead of deriving it by prefix. Nothing else changes.

**Re-verified**, this time on both builds — emitted bare (non-relative) import specifiers:

| build | main | this PR |
| --- | --- | --- |
| ESM | `node:path` | `node:path`, `i18next-fs-backend/cjs`, `i18next-http-middleware` |
| IIFE | *(none)* | *(none)* |

Numbers are unchanged from the original: `i18nBackend` chunk 153,593 → 788 raw, all chunks 1,247,449 → 1,094,644 raw (−12.2%) / 474,695 → 434,408 gzip (−8.5%), critical path unchanged.

PR #3122, opened 2026-08-25
feat(user-access-policy): arbitrary-header match dimension for access rules

HughParry:
Engine half of prosopo/captcha-private#2096 — "add header restriction to portal". Portal half: prosopo/captcha-private#4268.

Not ready for review.

## What

A generic header-matching dimension for access rules: a `Block` or `Restrict` rule can carry `headerName` / `headerValue` / `headerOperator` (`equals`, `contains`, `notEquals`, `notContains`, `notEqualsAny`, `notContainsAny`). Rebased onto `main`, on top of the browser match dimension (#3162).

## Condition is matched in code, not in Redis

Redis TAG queries can't express substring `contains`, can't apply a per-rule operator, and an **allow-list rule must fire on a request that doesn't have the header** (allow `x-app: ios` has to fire when there's no `x-app`). So the condition is carried on the rule and evaluated by `accessRuleHeaderMatches` against the raw request headers. An indexed `headerMatch` sentinel keeps header rules candidates through the JS-side scalar check; Redis returns them via the `no-user-scope` fall-through probe, the same route `os` and `browser` rules take. Only the header rules global to, or pinned to, the request's site come back — not the whole rule set.

`notEqualsAny` / `notContainsAny` carry a JSON value list (`encodeHeaderValueList`) and block unless the header matches one of them. They exist because an allow-list over several values of the same header can't be several single-value rules: `notEquals ios` + `notEquals android` each fire on the other's value and block everything. A malformed list makes the rule never fire rather than block everything.

## Request headers are a required argument

`getPrioritisedAccessRule` / `CaptchaManager.getPrioritisedAccessPolicies` take the raw request headers as a **required** parameter with no default. An earlier revision defaulted them to `{}` and wired only the block middleware and the verify-time hard-block check; the `/frictionless` policy and dedup lookups and the image / PoW / puzzle challenge endpoints ran blind, and because a negated operator treats a missing header as "does not match", every allow-list rule fired on every request at `/frictionless`. All seven lookups now pass headers via `normalizeHeadersForMatching`, and a unit test drives `getPrioritisedAccessRule` with an allow-list rule (matching header passes, absent header blocks, two requests sharing a cached candidate fetch each get their own verdict).

## Caching change

`getPrioritisedAccessRule` caches only the **candidate fetch** and ranks per request: `ruleApplies` consults the headers while the cache key is built from the user scope alone, so a cached ranked list would serve one request's header verdict to another. This also closes a latent gap where `os` was never part of `hardBlockCacheKey`.

## Pin for captcha-private

`pin/header-restriction` = these commits on top of `pin/2095-browser-restriction` (v3.8.4-based), so the private repo's lockfile stays as is until the next release; `main` here carries the dependabot bump (#3164) that the private lockfile doesn't yet.


HughParry:
Updated: added `notEqualsAny` / `notContainsAny`, whose `headerValue` carries a JSON value list (`encodeHeaderValueList`) and which block unless the request's header matches one of the listed values.

They exist because an allow-list over several values of the **same** header can't be expressed as several single-value rules — each fires independently, so `notEquals ios` + `notEquals android` blocks an iOS request (it trips the android rule) and an Android request alike, i.e. everything. That's the shape the ticket asks for: `os` matching `[windows, mac]`.

Deny-lists keep one rule per value. A list-shaped `headerValue` that isn't a usable list (empty, non-strings, malformed JSON) makes the rule never fire rather than block everything, matching the existing fail-safe for an unrecognised operator.

PR #3144, opened 2026-08-27
feat(provider,types): add a puzzle band to the frictionless flow

forgetso:
## What

`settings.frictionlessThreshold` becomes an object with two rungs instead of a single number:

```ts
frictionlessThreshold: {
  frictionlessPuzzleThreshold: 0.5,  // at or below -> PoW
  frictionlessImageThreshold: 1.0,   // at or above -> image; between -> puzzle
}
```

Scores at or below the puzzle rung still pass silently to PoW and scores at or above the image rung still get an image captcha. Everything in between — suspicious without being conclusive — now gets a puzzle instead of being lumped in with the worst traffic.

## Compatibility

The puzzle rung defaults to the value the setting already carried, so no site's silent-pass boundary moves. Putting both rungs on the same value opts out of the middle band and restores the previous two-outcome behaviour exactly.

A bare number is still accepted wherever the setting is read or parsed, and means what it always meant. `resolveFrictionlessThreshold` is the single place that knows how to read either shape, shared by the provider, the portal API and both migrations. That makes migration a background tidy-up rather than a cutover.

Unlike the puzzle rung, the image rung is not capped at 1 — the score it is compared against is a total that server-side penalties add to.

## Also here

- Image challenges on the score path are sized by how many signals fired rather than a fixed count, capped so one unusual trait can't buy an unreasonably long challenge.
- New `BOT_SCORE_PUZZLE_BAND` selection reason and `bot_score_puzzle_band` metric.
- A one-shot startup migration for provider client records, which are per-provider and so out of reach of a central migration. Awaited at boot but it never throws — the read path already understands the old shape. **Remove next release.**
- Cypress spec covering all three bands. Headless cypress can't produce a real score, so the score is supplied from the other side: a Restrict access rule with `frictionlessScore` and no `captchaType` bumps it and falls through to the decision machine.

## Testing

1564 unit tests pass. Two failures on this branch (`decryptPayload`, and the `dev/config/dist` bundle test) were verified identical on a clean tree.


forgetso:
Pushed a follow-up: **router-supplied image rounds are now clamped to the sitekey's `imageMaxRounds`.**

Every path that sizes an image challenge already respected that ceiling — the decision machine's gates, the access-policy dispatch, the traffic-filter dispatch, the configured-type short-circuit. A routing machine's `solvedImagesCount` was the one that didn't: `sendCaptcha` took it verbatim, and `RoutingMachineOutput` only bounds it as `int().positive()`, so a router could serve more rounds than the site had configured.

The ceiling now rides on the routing context. It stays optional because the dedup replay builds a context purely to ask which `captchaType` the router would pick and never creates a session from the answer — an absent ceiling must not be read as a ceiling of zero. Six regression tests cover the clamp, the pass-through, the puzzle→image downgrade, and that absent-ceiling case.

No-op against the routers in use today, which return counts well under the default of 32.

forgetso:
## Expected change in the challenge mix

On the default rungs (0.5 / 1.0), measured against a 24h sample of production sessions:

| | before | after |
|---|---|---|
| PoW | 89.8% | 89.8% (unchanged) |
| puzzle | 0% | 5.1% |
| image | 10.2% | 5.0% |

Those percentages are of the sessions that reach the score gate. Across the whole flow — image captchas also come from paths the ladder never touches, such as a missing token or a payload that failed to decrypt — the expected effect is roughly **a third fewer image captchas** and **around 40% more puzzles** on top of what routing machines already send.

PoW is unaffected: the lower rung is the value `frictionlessThreshold` already carried, so nobody who passes silently today stops passing.

Sites with a tuned threshold keep their own lower rung, so their split will differ from the above.

PR #3145, opened 2026-08-27
feat: Web Bot Auth verifier + authenticated frictionless flow

forgetso:
## Summary
- New `@prosopo/web-bot-auth` package: RFC 9421 HTTP Message Signatures verifier built on `@noble/curves/ed25519` (no Cloudflare dep). Parses Signature-Agent (bare-string + dictionary forms), resolves JWKS at `/.well-known/http-message-signatures-directory` with cache-control TTL, verifies Ed25519 over the RFC 9421 signature base.
- Frictionless authenticated fast-path in the provider: when a Web Bot Auth verified request matches an Allow policy at `/captcha/frictionless`, returns `captchaType: authenticated` with a session record (`serverChecked: false`, `agent: true`, `webBotAuthAgent`, IP frozen for verify-time binding). Skips decrypt / bot-score / decision machine.
- New `/client/authenticated/verify` router with mandatory IP-binding (`API.AUTHENTICATED_IP_REQUIRED` / `API.AUTHENTICATED_IP_MISMATCH`), single-use via `serverChecked`, `captchaType` gate against replaying other token types.
- `AccessPolicyType.Allow` and `CaptchaType.authenticated` added. `Session.agent` / `Session.webBotAuthAgent` fields for the "pre-verified pass" analytics filter.
- `ClientSessionId` + `email` support on the authenticated verify path: `submitAuthenticatedCaptchaVerify` on `ProviderApi`, `CaptchaType.authenticated` branch in `@prosopo/server.verifyProvider`, `verifyAuthenticatedSession` uses the shared `isClientSessionMismatch` helper for parity with pow/image/puzzle.
- `procaptcha-frictionless` gets `AuthenticatedBadge` + a dispatch branch for `CaptchaType.authenticated`.

## Test plan
- [ ] `npm run -w @prosopo/web-bot-auth test` — signature-base, structured-fields, Signature-Agent parsing, JWKS resolver, verify happy path
- [ ] `npm run -w @prosopo/provider build:tsc` — provider type-checks
- [ ] Frictionless entry with a signed request matching an Allow rule → `captchaType: authenticated`, session persisted with `agent: true`
- [ ] `/authenticated/verify` — missing IP → `API.AUTHENTICATED_IP_REQUIRED`; wrong IP → `API.AUTHENTICATED_IP_MISMATCH`; wrong clientSessionId → `API.CLIENT_SESSION_MISMATCH`; second use → `API.USER_ALREADY_VERIFIED`
- [ ] `@prosopo/server.isVerified` on an authenticated token dispatches to `submitAuthenticatedCaptchaVerify` (not the legacy image fallback)


forgetso:
Rebased onto `main` (was based on v3.7.27; conflicts in `provider/package.json`, `blacklistRequestInspector`, `checkTrafficFilter`, `frictionlessTasks`, `types/client/settings` and six `user-access-policy` files, all resolved in favour of keeping both sides — the local harshness tables were dropped for `@prosopo/captcha-severity`, which main moved them into). Added the missing `package-lock.json` entry for the new workspace and a changeset.

### End-to-end coverage

`packages/provider/src/tests/integration/webBotAuthAuthenticated.integration.test.ts` — 15 cases over real HTTP against a real provider, Mongo and Redis. The Web Bot Auth half signs with a generated Ed25519 keypair and serves the JWKS from a throwaway origin, so the directory fetch and signature verification are exercised for real rather than stubbed.

- entry: verified signer allowed → `captchaType: authenticated` (and the JWKS really was fetched); expired signature; Signature-Agent with no signature; non-signature qualifier (user agent); Block beats Allow; `deferToVerify` Allow ignored at entry; no matching rule
- verify: happy path; second redemption refused; missing IP; wrong IP; session still redeemable after a rejected verify; `clientSessionId` match and mismatch; non-authenticated token refused; token with no session id

### Three bugs it turned up, each of which broke the flow outright

1. **`ipMatchesSession` never matched.** It compared the operator's parsed IP against the session's composite halves with `===`. A session read back from Mongo carries BSON (`Decimal128`, or `Long` on pre-migration records), not the `bigint` the type claims, so every legitimate redemption was rejected as `API.AUTHENTICATED_IP_MISMATCH`. Both halves are now normalised first; an unparseable half fails closed rather than defaulting to `0n`, so garbage still cannot match garbage. Unit coverage added alongside.
2. **`serverChecked` was never written** onto the authenticated session, so "never set" and "consumed" were told apart only by an absence. Now written as `false` at issuance.
3. **`serverChecked` was missing from `SESSION_PROJECTION`.** Left out, the single-use check reads `undefined` and an authenticated token verifies an unlimited number of times. (Caught by the projection's own compile-time guard once the field was read.)

### One thing left as-is

`ClientSettingsSchema.allowAgents` is declared but never read — the fast path gates on a non-`deferToVerify` Allow rule matching the user scope, per the design the handler comment sets out. Left in place rather than removed or wired up, since deciding between those is your call. Worth noting before anything builds a portal toggle on it.

Local: full build, `build:tsc`, provider unit (1221) and the whole provider integration suite (15 files) all green.

PR #3151, opened 2026-08-28
fix(user-access-policy): fetch deferred rules at verify, not at request time

forgetso:
## Problem

`deferToVerify` rules are skipped by every request-time consumer and enforced by `checkForHardBlock` at verify. `findHardBlockPolicy` is written to accept a deferred rule of any policy type — its case (c) documents a deferred `Restrict` explicitly.

But its only caller fetches with `blockOnly`, which narrows the Redis candidate pool to `@type:{block}`:

```ts
// captchaManager.ts
const accessPolicies = await this.getPrioritisedAccessPolicies(..., { blockOnly: true });
return findHardBlockPolicy(accessPolicies);
```

So a deferred `Restrict` is never fetched and can never fire. Case (c) is unreachable. In production this leaves a population of deferred `Restrict` rules that silently do nothing.

The converse is also true at request time: a deferred `Block` matches `@type:{block}`, gets fetched by the middleware, and is then discarded in JS at `blacklistRequestInspector.ts:550` — work done for a rule that is never enforceable there.

## Fix

Index `deferToVerify` and select the hard-block pool per caller:

| caller | clause |
|---|---|
| verify (`checkForHardBlock`) | `(@type:{block})\|(@deferToVerify:{true})` |
| request time (middleware) | `@type:{block} -@deferToVerify:{true}` |

Deferred rules are now fetched exactly where they are enforced, and filtered out in Redis where they aren't. The JS-side `!p.deferToVerify` filter stays as a belt-and-braces guard.

`hardBlockCacheKey` takes the new flag so the middleware and verify lookups — same scope, different pools — can't share a cached verdict.

## Notes

- The union needs each side separately parenthesised. `@type:{block} | @deferToVerify:{true}` is a **RediSearch syntax error** across two fields; caught by testing against a live RediSearch rather than by inspection.
- Verified against live RediSearch that rules with no `deferToVerify` field survive the negation (the field is indexed `INDEXMISSING`), so plain Block rules are unaffected in the middleware pool.
- Adding an indexed field changes the index hash, so `createRedisIndex` drops and recreates once on startup, reindexing existing rules onto the new field.

## Verification

- `build:tsc` clean for `@prosopo/user-access-policy` and `@prosopo/provider`
- Unit tests for both pool clauses and the cache-key split
- New `blockingRuleShapes.unit.test.ts` characterises the two blocking shapes against the write-side schemas, including that a batch is rejected wholesale when one group is malformed
- `redisRulesStorage.integration.test.ts` passes in isolation; the 2 failures seen when the whole `redis/` directory runs together reproduce identically on a clean tree (pre-existing order-dependence with the benchmark test)


forgetso:
## Benchmark + a bug the benchmark caught

Benchmarked the two-field clauses against the single-field baseline. Added `deferredPoolClause.benchmark.ts` (standalone script, not picked up by CI) using the same population shape as `redisRulesReaderLoad.benchmark.integration.test.ts`.

**Population:** 21,300 rules — 17k ip-block, 1.3k cidr-block, 1k mixed, 2k deferred-restrict. Exact-IP probe, 50 warmup + 500 iterations, live RediSearch.

| clause | hits | mean | p50 | p95 | p99 |
|---|---|---|---|---|---|
| baseline `@type:{block}` | 1 | 0.105ms | 0.108 | 0.125 | 0.142 |
| middleware `@type:{block} -@deferToVerify:{true}` | 1 | 0.096ms | 0.093 | 0.116 | 0.135 |
| verify `(@type:{block} \| @deferToVerify:{true})` | 1 | 0.102ms | 0.102 | 0.113 | 0.139 |
| **misgrouped** `(@type:{block})\|(@deferToVerify:{true})` | **18800** | 0.760ms | 0.756 | 0.790 | 0.849 |

CIDR mask-range probe is the same story (0.094 / 0.091 / 0.108 vs 0.764ms).

**Both new clauses are latency-neutral against the baseline.** The middleware clause is marginally faster — the negation shrinks the posting list before the scope intersection.

### The bug

The union as first written, `(@type:{block})|(@deferToVerify:{true})`, is **wrong**. `|` binds looser than the implicit intersection, so composed with the probe it parsed as:

```
@type:{block}  OR  (@deferToVerify:{true} AND scope AND ip)
```

The left operand escaped the scope and IP filters and matched **every Block rule in the index** — 18,800 candidates instead of 1, and ~7× slower. At verify that means `checkForHardBlock` HGETALLs the entire Block population on every call, and `SERVER_SIDE_RANK_TOP_N` truncation then decides which rules survive.

It is now `(@type:{block} | @deferToVerify:{true})` — one group, valid under `DIALECT 2` which every caller uses (`redisAggregate.ts:46`).

Worth recording why it got through the first check: I validated against a 4-document fixture where the misgrouping happened to return the right rows. It is invisible below population scale. The benchmark now runs the misgrouped clause as a control, and a unit test pins the grouping.

### Also

The `#policy/*` import alias resolves to `./dist/*`, so unit tests run against compiled output — earlier runs were partly against stale `dist`. Re-verified after a clean rebuild:

- `@prosopo/user-access-policy` unit — 109 pass
- `redisRulesStorage.integration.test.ts` (isolated) — 36 pass
- `hardBlockVerdictCache` + `blacklistRequestInspector` unit — 62 pass

PR #3162, opened 2026-09-01
feat(user-access-policy): browser match dimension for access rules

HughParry:


HughParry:
Not ready for review — still watching CI.

HughParry:
Verified end to end on a local stack (provider + portal + demo, Mongo/Redis in docker):

- A `browser: firefox` Block rule seeded in Mongo synced through the job pipeline to Redis and fanned out across all four demo site keys.
- Against the provider's frictionless endpoint on the demo site key: a Firefox User-Agent gets **403 Forbidden** at the block middleware, while Chrome and Safari pass it and reach the handler.

Also corrects a claim in the changeset: I'd written that the Redis index needs rebuilding on deploy. It does not — `createRedisIndex` hashes the index definition and drops/recreates when it changes. Confirmed locally: after the schema change the index came back carrying the `browser` TAG field and the stored hash moved to match. Changeset updated.

PR #3180, opened 2026-09-02
feat(types,provider): share the captcha-type severity ranking

forgetso:
Superseded by #3182, which places the shared ranking in a dedicated zero-dependency package rather than in `@prosopo/types`, and covers all three call sites instead of two.

Closed unmerged. See #3182 for the change that landed.


forgetso:
Superseded by #3182, which places the shared ranking in a dedicated zero-dependency package rather than in `@prosopo/types`, and covers all three call sites instead of two — including an ordering bug in the access-rule harshness scale.

PR #3245, opened 2026-09-11
feat(procaptcha): keyboard and screen reader access for the challenges

forgetso:
## What was wrong

A client asked how our puzzle presents to people using a screen reader. The answer was: it doesn't.

The draggable piece was a plain `div`. You could not reach it with the Tab key, it had no name and no role, and nothing was announced when the puzzle appeared. So a screen reader user could tick "I am human", get a silent overlay, and be stuck with no way forward and no way back except Escape, which nothing told them about. The image captcha's tiles had exactly the same problem.

## What changed

**Puzzle**
- The piece can be focused and moved with the arrow keys. Hold shift for smaller steps, Home puts it back at the start, Enter or Space submits it. It has a visible focus ring.
- A live region reads out where the piece is (as a percentage across and down), that an answer is being checked, and that a failed go has been replaced with a fresh puzzle.
- Its on-screen text was hardcoded English. It now goes through the locale package and is translated into all 32 languages.

**Image captcha**
- The tiles are buttons instead of clickable `div`s, so they can be tabbed to and picked with Enter or Space, and they report whether they are selected rather than only looking selected.

**Both**
- The challenge now opens as a proper dialog: it takes focus when it opens, keeps Tab inside itself, and gives focus back to the checkbox when it closes.

**All three captcha types**
- The spinner that replaces the checkbox while a check runs used to drop focus to the top of the page without saying why. It now takes focus in the checkbox's place, names itself, and hands focus back when the check finishes.

## What this does not fix

A visual puzzle still cannot be solved by someone who cannot see it. The widget is deliberately never told where the target is, so there is nothing it could describe. Sites that need a challenge a blind user can complete should use the pow captcha type, which needs no interaction beyond the checkbox. PoW was already accessible apart from the spinner issue above.

## Test coverage

775 tests pass across the seven affected packages. New tests cover: arrow-key movement in all four directions, the finer shift step, clamping at the board edge, Home, Enter and Space submitting with the right trail, rejection of key events no user made, the dialog's name and modal flag, focus landing on the piece and returning to the opener, the Tab trap, and each of the announcements. On the image side: the tiles being buttons and reporting their pressed state.

## Note on the trade-off

The piece's ARIA attributes and CSS class are a stable selector in production builds, which the existing `data-cy` deliberately is not (see the comment in `PuzzleCanvas.tsx`). A scripted solver could already find that element — it is the only draggable thing on the surface — whereas without them a keyboard or screen reader user cannot find it at all. Flagging it rather than deciding it quietly.

Also worth a second opinion before release: a keyboard solve produces a perfectly linear, evenly spaced event trail. I could not find a decision machine rule that consumes `puzzleEvents`, but whoever owns scoring should confirm, otherwise accessible users just get failed a different way.


forgetso:
Follow-up for the keyboard-solve scoring question raised in review: prosopo/captcha-private#4539 (private repo).

Short version: we cannot detect a screen reader and should not try — the target is "is this a human", as it always was. The issue covers marking the input modality so keyboard solves are countable, capturing key-level timing (auto-repeat cadence is a decent biometric), and the constraint that the accessible path must not end up harder to pass than the mouse path.

PR #3298, opened 2026-09-19
test: cypress baseline control (do not merge)

forgetso:
Control run only — establishes whether `invisible.cy.ts` is green on an unmodified main, to decide whether the failure on #3297 is caused by that change or pre-existing. Not for merge; will be closed once the answer is in.


forgetso:
Control run complete — established that the cypress failures on #3297 were flakes rather than caused by the decoder change (different specs failed on each run; a re-run went green). Closing; not for merge.

PR #3312, opened 2026-09-21
fix(widget): stop the image grid wrapping to two columns on a phone

forgetso:
Fixes the image challenge showing a 2×5 grid instead of 3×3 on Android, with the bottom images off-screen and unreachable.

## What was wrong

Measured on the reported device (Samsung A52s, Brave, 384×718 CSS px, attached over adb):

- the image grid is **345.578px** wide
- each tile asked for `calc(33.333% - 5.33px)` at an 8px gap
- three tiles plus the two gaps between them = **345.584px**

That is **0.0065px too wide**, so the browser wrapped the third tile onto the next line. The challenge drew 2 columns and 5 rows instead of 3 and 3, the grid went from ~369px to **1076px** tall, and 3 of the 9 images ended up below the 718px viewport.

Overriding just the tile width on the live page, to what this PR emits, put it back to 3 columns and 369px.

## The cause

This came in with the dialog randomisation (#3306), which replaced a fixed `calc(33.333% - 10px)` with a width derived from the randomised gap. The old value subtracted a **whole** gap per tile where only two thirds is needed, so it carried ~10px of slack and always fitted. Deriving it exactly removed the slack, and `Math.round((200 * gap) / 3) / 100` then rounded the two thirds *down* for some gaps, tipping three tiles over the edge.

Which gaps break is pure arithmetic — 8, 11 and 14 round down and overflow; 9, 10, 12, 13 and 15 do not. So roughly a third of page loads, which is why it looked intermittent.

## The fix

`calc((100% - 2*gap px) / 3)` — both the subtraction and the division left to the browser, so there is nothing to round and three tiles plus their gaps come to exactly the row width at every gap.

## Also here

The panel is `touch-action: pan-y` instead of `touch-action: none`. It is an `overflow-y: auto` box, so on a phone it is the thing a finger drags when the images do not fit; `none` tells the browser not to pan it at all, which is what turned "some images are off-screen" into "some images cannot be reached". `overscroll-behavior: none` is what stops the page behind scrolling, and it stays.

That one is **not** from the randomisation — it predates it, arriving with the React→vanilla rewrite (#2991), and the React component before that had the same declaration. With the grid fix the panel no longer overflows in portrait anyway, so this is belt-and-braces for landscape and smaller screens.

## Test coverage

A new test works out where three tiles and two gaps land for every gap the challenge draws (6–15), against a set of panel widths including the A52s' 345.5777893066406. It fails on the old expression with exactly the numbers measured on the device:

```
gap 8 at width 345.5777893066406 used 345.58433352874755:
expected 345.58433352874755 to be less than or equal to 345.5777893066406
```

Full suites pass: `procaptcha-common` 338, `procaptcha-react` 207.

## What is not verified

The `touch-action` half is argued from the computed style I measured on the device plus the CSS spec, not from a demonstrated drag — two headless harness attempts scrolled zero with *either* value, so the harness was broken and proved nothing. Worth a look on a real phone once this is on staging.


forgetso:
Verified on staging on the reported hardware.

This build was deployed to `staging-js.prosopo.io` with `--local` (bundle only) and confirmed live by crawling all 23 chunks: the tile width is now `calc((100% - ${2 * gap}px) / 3)` and the image panel is `touchAction: "pan-y"`. Entry and the chunk carrying the fix are byte-identical to the local build.

Chris confirms the image challenge now renders correctly on the Samsung A52s that reproduced the 2×5 grid, and on iPhone.

Still open, and deliberately not in this PR: the two-tap checkbox on iOS. That is the hover state layer on the checkbox, predates the randomisation work, and needs its own change.

